PR #44, opened 2021-06-15
Switch project to using MySQL in dev instead of SQLLITE

### LINKED ISSUE #41: Switch project to using MySQL in dev instead of SQLLite
Will need to switch the Docker Compose and the Django settings.

Example where we are doing this: https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/docker-compose.yml

danie1zhang:
This PR modifies modifies `docker-compose` and `backend/settings.py` to switch the project to using MySQL in dev instead of SQLLITE. It adds a `mysql` service to `docker-compose`, adds a configuration approach by adding a container environment using the `environment` block in `docker-compose`, and sets up the database connection in Django by adding a `mysql` database connection in `backend/settings.py`. This PR aims to resolve #41.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
 
   mysql:
     image: mysql:5.7
-    restart: on-failure
+    restart: unless-stopped
     command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci']
     environment:
       - MYSQL_ROOT_PASSWORD=cae_root_pw
@@ -14,7 +14,7 @@ services:
       - MYSQL_USER=cae_user
       - MYSQL_PASSWORD=cae_pw
     ports:
-      - "5306:3306"
+      - "6306:3306"
     volumes:
       - ./.data/mysql:/var/lib/mysql:delegated
     container_name: canvas_app_explorer_mysql
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,23 @@
 version: "3.9"
    
 services:
+
+  mysql:
+    image: mysql:5.7
+    restart: on-failure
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,6 +1,23 @@
...
+    command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci']
+    environment:
+      - MYSQL_ROOT_PASSWORD=cae_root_pw
+      - MYSQL_HOST=canvas_app_explorer_mysql
+      - MYSQL_PORT=3306
+      - MYSQL_DATABASE=canvas_app_explorer_local
+      - MYSQL_USER=cae_user
+      - MYSQL_PASSWORD=cae_pw
+    ports:
+      - "5306:3306"
```
jonespm, line 17: I'd change this 5306 to a different port, like maybe even 6306 is fine. This is really just so it doesn't conflit with our other local databases.

docker-compose.yml
```diff
@@ -1,6 +1,23 @@
 version: "3.9"
    
 services:
+
+  mysql:
+    image: mysql:5.7
+    restart: on-failure
```
jonespm, line 7: We've wanted to change this `restart: on-failure` to `restart: unless-stopped`. I'd make this change here for this new project. https://docs.docker.com/config/containers/start-containers-automatically/

```suggestion
    restart: unless-stopped
```

jonespm (changes requested):
A few minor changes but this looks fine.

We may want to change this from `os.getenv` to something else in the future but this is fine for now.

jonespm (approved):
I've resolved my comments as you incorporated this and will merge it. Thanks!

Issue #45, opened 2021-06-15
Create the Django Data Model


danie1zhang:

jonespm:
This is the same as #46 I believe, just close that one when this one is resolved.

PR #46, opened 2021-06-18
Define Django Data Models and Register Them

danie1zhang:
This PR modifies `backend/settings.py` and creates new files `backend/models.py` and `backend/admin.py` to define Django Data Models and register them. So far, this PR has defined the data models in `backend/models.py`, added `backend` to `INSTALLED_APPS` in `backend/settings.py`, and registered these data models in `backend/admin.py`. This PR has not been able to migrate the new app in `INSTALLED_APPS` because the docker has some problems related to mysql. It attempts to address issue #45.


jonespm:
Thanks, you can assign me as a review on these pull requests.

backend/admin.py
```diff
@@ -0,0 +1,10 @@
+from django.contrib import admin
+from canvas-app-explorer.backend.models import LtiTool, CanvasPlacement
```
jonespm, line 2: This import should just be

```suggestion
from backend.models import LtiTool, CanvasPlacement

```

jonespm:
Hi Daniel,

I added another commit to this, moving everything into the canvas_app_explorer app created with the start command. I also fixed up manage so it adds this. 

I also ran
`docker exec -it canvas_app_explorer ./manage.py makemigrations`

To create the migrations file. I'm seeing everything working okay. There's a new error with a CSS problem with admin site, but we can fix that separately.

jonespm:
I'm reviewing the model. The only thing I'd change is fix the mispelling of 

`canvas_placmenet`

And set that as a ForeignKey.

jonespm (approved):

danie1zhang:
These changes look good to me, I am going to squash and merge.

PR #53, opened 2021-06-22
Fixes #50 - Admin page isn't able to load static css files.

### LINKED ISSUE #50: Admin page isn't able to load static css files
When logging into http://localhost:5000/admin/ the css isn't working and it's not able to get some files from static. This is probably either because these aren't being collected or the path isn't exposed correctly. Will look into fixing this. 

image

jonespm:
Move django static files to a different path (django_static) and only add the frontend in development mode.

This will need to change a bit as we move to production but I feel like I see what's going on here now.


danie1zhang (approved):
I just ran these changes and both admin and the original webpage worked for me, so I'm going to approve.

PR #56, opened 2021-06-22
Fixes #55 Adding a __str__ method to each model class so it will disp…

### LINKED ISSUE #55: In Canvas Admin, display the name rather than the object id in the list
Rather than show the object id, we should show the name. This is in the Django admin and easy to tell the difference. 

image

jonespm:
…lay the name


danie1zhang (approved):
Code still runs properly, and changes look good, so I will approve

Issue #57, opened 2021-06-23
Allow uploading image files through the admin site 


jonespm:
The admin site doesn't show any way to preview or upload image files. We might need a package like this one to make that happen: https://github.com/victor-o-silva/db_file_storage

Or a widget as described here: https://stackoverflow.com/questions/46692520/how-to-upload-files-into-binaryfield-using-filefield-widget-in-django-admin

This should also return the value as base64 from DRF once that is implemented.

PR #61, opened 2021-06-25
Fixes #60 Consumes REST API using React to generate Canvas App Components

### LINKED ISSUE #60: Generate components by consuming REST API

danie1zhang:
Starting point for addressing issue #60. Fetches JSON file from REST API, and using JSON data to create new ProductCards with custom information.


jonespm:
We talked on the meeting that this should
* Do a for each on the return from the API call and just use that rather than a fixed value
* Include the headers and footers from the PlasmicHome component
* Do a few other touchups on the cards

It's possible the images won't be working here yet but we can fix that with #57

danie1zhang:
@jonespm Implemented all the necessary code in Home.jsx, but can't get the json to work. I was wondering if you had any idea why. When I use async await, there isn't even a GET request in the terminal for /api/lti_tools/ and the webpage doesn't load, and when I remove async and await, a GET /api/lti_tools is shown, but the webpage still doesn't load.

jonespm (approved):
I think this is good, there is more cleanup we could make on this Home component but we'll make those in future issues.

PR #63, opened 2021-07-02
Fixed #57 Adding package and migrations to store image files in the database

### LINKED ISSUE #57: Allow uploading image files through the admin site

jonespm:
Overriding the url method to get images instead of download them


danie1zhang (approved):
I tried out the code, and now I'm able to upload images as admin, so it looks good to me. I am going to merge.

PR #74, opened 2021-07-15
Support for LTI launches Fixes #1

### LINKED ISSUE #1: Integrate with PyLTI to Launch 1.3 from Canvas

jonespm:
I wrote up this guide in wiki to test here: https://github.com/tl-its-umich-edu/canvas-app-explorer/wiki/Configuring-LTI-1.3

Let me know if you have any feedback. This looks pretty good to me. We *might* need to switch to the json configuration before going to production for the cache to work correctly.


frontend/server.js
```diff
@@ -2,13 +2,17 @@ const webpackDevServer = require('webpack-dev-server');
 const webpack = require('webpack');
 
 const config = require('./webpack.config.js');
+const publicPath = 'https://958ba9a1711d.ngrok.io/dist/'
```
zqian, line 5: Is this a placeholder for ngrok url?

backend/settings.py
```diff
@@ -134,4 +135,27 @@
...
+
+# TODO: Switch this to CSP for additional security
+X_FRAME_OPTIONS = 'ALLOWALL'
+
+# So request works over the proxy
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+
+DB_CACHE_CONFIGS = os.getenv('DB_CACHE_CONFIGS',
+                           {'CACHE_TTL': 600, 'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
+                            'LOCATION': 'django_myla_cache',
```
zqian, line 148: still says django_myla_cache here

backend/settings.py
```diff
@@ -134,4 +135,27 @@
...
+                            'CACHE_KEY_PREFIX': 'myla',
```
zqian, line 149: and prefix is still myla

zqian (commented):
I have tested the code and made some minor changes to the installation wiki page

backend/settings.py
jonespm: Thanks! I copied some of this from myla, I'll update this name.

frontend/server.js
jonespm: Hmm, I think I just left that in there and it should use the configured value. I'll remove it. Good catch!

jonespm: I'm trying to figure out why I had this in there in the first place. There were some issues getting this to work with ngrok. 

This configuration and frontend server is only used in local development, it should be compiled down to one server (just the backend) in production.

zqian (approved):

frontend/server.js
jonespm: I think I remember why I had this now. The backend code actually uses some of these variables from the frontend to figure out how to serve up the content. I'd temporarily hardcoded this but figured it out more and fixed it.

jonespm:
I'd like to fix #80 before this one, otherwise this causes warnings after running this migration that there are more migrations to run. I should have that fixed tomorrow and we can get these both in.

Issue #16, opened 2021-06-01
Implement Learn More Window


danie1zhang:

danie1zhang:
Learn More Card Works

Issue #14, opened 2021-06-01
Instructor should be able to add tools to their course navigation


danie1zhang:

zqian:
please track further development of add/remove LTI tool feature in the #21

PR #75, opened 2021-07-19
Implementing Learn More Button and Add Remove Button.

danie1zhang:
This pull request addresses issue #14 and #16. Currently able to have add/remove work for all the buttons at once but not each button/tool individually (if user clicks add, it changes for all tools, and if user clicks remove it changes for all tools). Still working on Learn More, it follows a similar idea.

### DIFF
--- a/frontend/components/Home.jsx
+++ b/frontend/components/Home.jsx
@@ -28,6 +28,8 @@ export const Home__VariantProps = new Array();
 export const Home__ArgProps = new Array();
 
 function Home__RenderFunc(props,ref) {  
+    const [addedTools, setAddedTools] = useState([]); // each tool has one entry in array
+    const [learnMoreActive, setLearnMoreActive] = useState(false) // each tool has one entry in array
     const [tools, setTools] = useState(null);
 
     useEffect(async () => {
@@ -35,14 +37,14 @@ function Home__RenderFunc(props,ref) {
         const response = await fetch(url);
         const data = await response.json();
         setTools(data)
+        setAddedTools(Array(Object.keys(data).length).fill(false))
     }, []);
 
-    console.log("Tools");
-    console.log(tools);
 	const { variants, args, overrides, forNode, dataFetches } = props;
 	const globalVariants = ensureGlobalVariants({
 		screen: useScreenVariants()
 	});
+
 
 	return (
 		<React.Fragment>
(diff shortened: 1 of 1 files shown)


frontend/components/Home.jsx
```diff
@@ -22,28 +22,26 @@ import "@plasmicapp/react-web/lib/plasmic.css";
...
+    const [addedTools, setAddedTools] = useState([]); // each tool has one entry in array
+    const [learnMoreActive, setLearnMoreActive] = useState([]) // each tool has one entry in array
     const [tools, setTools] = useState(null);
 
     useEffect(async () => {
         const url = "http://localhost:5000/api/lti_tools/";
         const response = await fetch(url);
         const data = await response.json();
         setTools(data)
+        setAddedTools(Array(Object.keys(data).length + 2).fill(false))
```
jonespm, line 37: I don't think we really need to initialize this and can just leave it undefined and set the values that are enabled to true. This may also change to something like an object instead of an array.

Ultimately we'll be using the [tabs API in Canvas](https://canvas.instructure.com/doc/api/tabs.html) to get the tools that are placed and map them to the tools in our system.

That API returns a tool like
```
  {
    "id": "context_external_tool_12345",
    "html_url": "/courses/123456/external_tools/12345",
    "full_url": "https://umich.instructure.com/courses/123456/external_tools/12345",
    "position": 18,
    "visibility": "public",
    "hidden": true,
    "label": "Zoom",
    "type": "external",
    "url": "https://umich.instructure.com/api/v1/courses/123456/external_tools/sessionless_launch?id=12345&launch_type=course_navigation"
  },
```

So maybe we'll have something here like

toolsHidden: {
  "12345": false,
  "23456": true
}

Probably something like this idea to get it to update.
https://stackoverflow.com/a/54150873/3708872

It's fine to push that off to a later task.

frontend/components/Home.jsx
```diff
@@ -139,89 +133,131 @@ function Home__RenderFunc(props,ref) {
...
+                                    {learnMoreActive[tool.id] === false ?
+                                        <ProductCard
+                                            data-plasmic-name={tool.name+"Card"}
+                                            data-plasmic-override={overrides.zoomCard}
+                                            addRemoveSlot={(
+                                                <div>
+                                                    {addedTools[tool.id] === false ?  
+                                                        <AddRemoveButton
+                                                            onClick={(e) => {
+                                                                console.log(HELLOOO)
```
jonespm, line 146: Please remove these log messages if they're not needed.

frontend/components/Home.jsx
```diff
@@ -139,89 +133,131 @@ function Home__RenderFunc(props,ref) {
...
+                                                            }}
+                                                            className={classNames(
+                                                                "__wab_instance",
+                                                                // "Home__addRemoveButton___3D93M" // I don't see any changes in formatting with this line
+                                                            )}
+                                                        />
+                                                    :
+                                                        <AddRemoveButton
+                                                            onClick={(e) => {
+                                                                console.log(YOOOOOOOOO)
```
jonespm, line 160: Also remove this one

jonespm (changes requested):
Just a few minor comments and cleaning up some logs and I'll merge.

I think we could, rather than having two AddRemoveButtons just put the logic around the removeToolFromSite slot fill (currently on like 170). The other logic is identical in both methods.

jonespm (approved):

PR #83, opened 2021-07-29
Fixes #80 Changing default back to AutoField and setting BigAutoField for our app

### LINKED ISSUE #80: Avoid missing migration warnings about BigAutoField on packages
We're getting some warnings that migrations are missing on some packages because they have a different AutoField.

It looks like we should just change the default back to AutoField and make our project BigAutoField to avoid these warnings.

https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys

### LINKED ISSUE #81: Switch nodejs install over to nvm
I was having some issues installing node with the steps we have, going to switch it over to nvm based on this:
https://stackoverflow.com/a/57546198/3708872

Looks like less dependencies and can pick a a more specific version.

jonespm:


backend/canvas_app_explorer/apps.py
```diff
@@ -2,4 +2,5 @@
 
 
 class CanvasAppExplorerConfig(AppConfig):
+    default_auto_field = 'django.db.models.BigAutoField'
```
zqian, line 5: Add a comment here for why use BigAutoField, instead of the default AutoField.

zqian (approved):
Tested and the app ran successfully after the change. 

I suggested to add a comment in the apps.py as the reason of using BigAutoField instead of the default AutoField.

backend/canvas_app_explorer/apps.py
jonespm: We **could** have just used AutoField but that would have required re-doing the conversions. Prior to Django 3.2 these were AutoField but you had to be explicit about it.

jonespm: I think the reason they were BigAutoField in the first place was that was what they were in the example and I didn't notice until having as the project default caused a problem. :/ It doesn't make a huge difference either way. This is just one way of specifying the default on a package rather than application level now.

jonespm: Maybe I'll just check on this with @danie1zhang and we can just re-do the migration from the start since we haven't went to production.

jonespm: I just added a migration back to AutoField. This is int(11) in the database which will be enough for all of our ids.

PR #88, opened 2021-08-06
Comment out ngrok from sample config

jonespm:


jonespm:
@danie1zhang confirmed that this is working, going to commit.

PR #94, opened 2021-09-08
Fixes #91 - Create separate dockerfile to be used with openshift

### LINKED ISSUE #91: Create separate Dockerfile to be used with Openshift
- Will need to reference the Openshift images
`FROM docker-registry.default.svc:5000/openshift/python:3.8-slim`

- Will need to collectstatic and only run the Gunicorn server (not the frontend)
  - Might need another supervisor file or just run Gunicorn directly.

jonespm:
This is still a work in progress, working to get a separate docker file setup for Openshift testing where everything is compiled and running in Guincorn. Will probably need a separate docker-compose. 

- [x] Create separate Dockerfile to test Openshift config
- [x] Document how to use docker compose to test
- [x] Customize startup scripts to start in development mode

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -5,6 +5,7 @@
 # python -c 'from django.core.management.utils import get_random_secret_key; print(get_random_secret_key())'
 # DJANGO_SECRET_KEY=<some secret key, randomized by default>
 
+##### Optional configuration for defining hosts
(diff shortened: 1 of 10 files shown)


danie1zhang (approved):
Ran the updated code and it works well for me. New additions to code looks good as well.

PR #95, opened 2021-09-20
Removing as app as it's causing issues with openshift

jonespm:
No issue, but this is simple. 

Openshift is giving an error with "as app" in here so just remove it. Works fine either way here locally.


jonespm:
I tested this and it's working, just going to merge it.

PR #97, opened 2021-10-06
Fixes #96 - Adds Django Rich Text Editor Quill 

### LINKED ISSUE #96: Implement Rich Text Editor for Django to allow for links
Here was a blog of different editors.

https://blog.devgenius.io/best-free-wysiwyg-editor-python-django-admin-panel-integration-d9cb30da1dba

We tried out Quill in #97 and it seems okay but it saves it's data in a custom format and can't convert existing data. Probably not a big deal for us to start fresh but I don't love proprietary data. Will briefly look into a conversion. 

All of these except the Aloha Editor would be license compatible.

danie1zhang:


jonespm:
Closing for #103

PR #103, opened 2021-10-20
FIxes #96 Second Try, Rich Text Editor using tinymce

### LINKED ISSUE #96: Implement Rich Text Editor for Django to allow for links

danie1zhang:


jonespm:
This looks good!

I did a minor change to the settings to add a comma in the list and also provide a slightly different default config that adds the menu bar in which I think is useful. I also added a migration which I don't believe does anything to the data field. Will double check this but looks okay.

I filed an issue for dark mode detection in the repository, but everything else looked better than the other plugin to me.

jonespm:
@danie1zhang Is this ready to move out of draft? I think we're good to merge this and close the other one.

jonespm (approved):

PR #108, opened 2021-12-01
Fixes #105 Tiny MCE causes text on cards to become HTML code

### LINKED ISSUE #105: TinyMCE causes Text on Cards to become HTML code
image

jonespm:
Okay, I think this fixes all of the fields.

I tested by adding color and also links. 

Learn More side of the card seems to have a large gap between description and privacy agreement, but this exists before this change so seems separate.


jonespm:
This will conflict with #106 so maybe that should be merged first.

PR #134, opened 2022-04-11
Limit `LtiTool.short_description` length to 120 characters (#128)

### LINKED ISSUE #128: Limit short description to be 120 characters
The short description should be limited to 120 characters, otherwise it goes to 4 rows and takes up too much space on the card.

ssciolla:
The PR aims to resolve #128.


ssciolla:
> I was looking and [TextField ](https://docs.djangoproject.com/en/4.0/ref/models/fields/#textfield) (which HTMLField extends) doesn't change the database so there wasn't anything different after the conversion. That's not optimal but seems fine.
> 
> Sounds like you'd have to use some MySQL specific solution to change this to a [SizedTextField ](https://django-mysql.readthedocs.io/en/latest/model_fields/resizable_text_binary_fields.html?highlight=mediumtext#django_mysql.models.SizedTextField)(TINYTEXT). Probably not worth it for this app, but interesting.

@jonespm, so you're saying it's just Django enforcing the length, and MySQL wouldn't? Interesting. Should I worry about this any further or just merge?

jonespm:
I'd just merge it. I think we'd if this was deployed externally we'd want to add a change to also limit number of characters returned from the API, or actually limit on the database. But we can do this manually if there's any that are too long already.

ssciolla:
Looks like we could use a validator to tighten it up at the model level; still wouldn't be enforced at the database level though. https://stackoverflow.com/a/15845514

ssciolla:
Actually upon further inspection, I'm not sure a validator would do anything else for us (specifying `max_length` may already use one). It is validating in the admin interface, so that's something. I'll just merge for now.

PR #138, opened 2022-04-16
Update `node` and `nvm` versions (#135)

### LINKED ISSUE #135: Adopt Node 16
We're moving our other Node projects to version 16, which is the current LTS, so we should do that here as well.

(Admittedly, a big reason for my wanting to make this change is the lock file version changes with the associated `npm` version updates, and I'd like to not have to switch back and forth between 14 and 16 when installing packages for different projects.)

ssciolla:
The PR aims to resolve #135.

Steps taken:
1) Update versions for `node` and `nvm`
2) Do a local install with `npm` 8, `node` 16 to update `package-lock.json`
3) Run `npm audit fix`

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -16,8 +16,8 @@ RUN pip install --no-cache-dir -r requirements.txt
 
 WORKDIR /code/frontend
 
-ENV NODE_VERSION=14.17.4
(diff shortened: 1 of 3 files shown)


jonespm (approved):
Looks good! I was installing it like this to try and avoid the multi-stage build that MyLA has for now, especially for the Openshift build. This may change now after we do a release and use a GCR for storing the intermediary containers. Good to stay updated though!

PR #143, opened 2022-04-18
Refactor `Home.jsx`, `ProductCard.jsx` (#142)

### LINKED ISSUE #142: Streamline `Home.jsx` by moving styles in Plasmic, giving some responsibilities to `ProductCard.tsx`
Currently, the `Home.jsx` wrapper seems to be mostly modeled after `PlasmicHome.jsx`, including its use of `makeNodeComponent` and the frequent use of class strings. The extended JSX tree this creates hampers readability, and the approach seems to be sidestepping some of the benefits of Plasmic, which is supposed to be handling the UI stuff like CSS. I think we can pare this down significantly, first in Plasmic Studio by localizing styles to certain components or elements and using image slots, and second by giving responsibility for mapping tool data and the learn more button to the `ProductCard.jsx` wrapper.

Hopefully the benefits and actions taken will make more sense by looking at the code. Part of the work has already been done in #141.

ssciolla:
This PR aims to resolve #142.


ssciolla:
Okay, I think this is ready for review. I made some additional small Plasmic Studio changes, but they shouldn't have any effect on this. Let me know using comments if you have any questions.

frontend/components/ProductCard.jsx
```diff
@@ -18,7 +21,59 @@ function ProductCard_(props, ref) {
...
+        title={tool.name}
+        description={<span dangerouslySetInnerHTML={{ __html: tool.short_description }} />}
+        learnMoreSlot={
+          <TitleLearnMoreButton
+            onClick={(e) => {
+              e.preventDefault();
+              setLearnMoreActive(true)
+            }}
+          >
+            {tool.name}
```
ssciolla, line 47: I find it kinda confusing that that we are using the name of the tool for the button. Was that something recommended by the UX team? I'd probably add a header for this variant with the tool name and then keep this as "Find Out More"

jonespm: We've been back and forth about this button. Looking at the latest UX review, this needs to be changed but it could be done in a new issue.

Change the blue button text
* Currently, the idea was to change it to “Find Out More,” with screen reader text saying “Find out more about X”
* One other suggestion from the DUX team is “More Details” with screen reader text saying “More X Details”
* Overall, this makes for more succinct button text

We need to test how the screenreader reads the content.

jonespm (approved):
Looks good. I'd create a new issue since we still need to address and test the text of the button.

PR #151, opened 2022-04-21
Add customizable alt text for images (#144)

### LINKED ISSUE #144: Add a way to provide alt text for uploaded images
The admin UI allows for some images (like screenshot and logo) to be uploaded. We should also provide a small text area somehow to allow for alt text for each of these.

ssciolla:
The PR aims to resolve #144. (Plasmic changes have already been made.)


jonespm:
Tested this out and looks good

One suggestion is to make this change to admin.py

Then the image and the alt text are on the same line in the admin UI

image
verses
image

Easiest way I could see to do it.

I saw an [app for adding alt text ](https://github.com/unplugstudio/django-alttext) to any FileField and what we have seems simple enough for now since we've just got the 2 images.

```
diff --git a/backend/canvas_app_explorer/admin.py b/backend/canvas_app_explorer/admin.py
index 44f9761..73fa64b 100644
--- a/backend/canvas_app_explorer/admin.py
+++ b/backend/canvas_app_explorer/admin.py
@@ -4,7 +4,17 @@ from django.contrib import admin
 from backend.canvas_app_explorer.models import LtiTool, CanvasPlacement

 class LtiToolAdmin(admin.ModelAdmin):
-    pass
+    fields = (
+        'name',
+        ('logo_image','logo_image_alt_text'),
+        ('main_image','main_image_alt_text'),
+        'short_description',
+        'long_description',
+        'privacy_agreement',
+        'support_resources',
+        'canvas_placement'
+    )
+
 admin.site.register(LtiTool, LtiToolAdmin)

 class CanvasPlacementAdmin(admin.ModelAdmin):
```

jonespm (approved):
I tested this out and it looks good. Just left a comment, can apply it or not. I think it probably is fine either way.

ssciolla:
> One suggestion is to make this change to admin.py
> 
> Then the image and the alt text are on the same line in the admin UI

Sure, done, that seems like a nice change. https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/151/commits/9e9c4317e767684d0343a18f352c5d4d2e26c8b8

Not totally convinced by the library, but maybe I'd have to try it out. Maybe if application was more about images, or we used template tags. It did occur to me that the data modeling is clunky (alt text and image path could be part of an image model, which could then be related to `LtiTool` M:1), but that's probably unnecessary, right?

ssciolla:
Gonna merge for now, thanks for the review.

PR #158, opened 2022-05-06
Adopt Material UI (#157)

### LINKED ISSUE #157: Replace Material UI with Plasmic
After some discussion, we decided it makes sense going forward to switch to using Material UI as a component library and CSS framework over Plasmic. Plasmic offered some nice features and was good for rapid protoptying, but we weren't quite happy yet with the UI appearance, felt it was challenging to keep things in sync, and saw that the frontend codebase was much more compact more with Material UI. This switch would/will also support maintenance, as two of our other tools use Material UI.

ssciolla:
This PR aims to resolve #157. It includes component and style solution replacements to Plasmic using `@mui` (Material UI version 5) and updates to `react` and `react-dom` (latest version of `17.02`).


jonespm (approved):
I've been testing this locally and it seems really nice. I don't have any comments to make about anything here, I'd say it's good to go when you think it is. 

I created a branch as main-plasmic to capture the old changes but don't think we'll need to go back. There will probably be some further cleanup and improvements we can make after this too.

backend/urls.py
```diff
@@ -44,21 +42,19 @@
...
+        re_path(r'^static/(?P<path>.*)$',
             view = static.serve, 
             kwargs = {
-                'document_root': os.path.join(settings.BASE_DIR, 'frontend/public/plasmic'), 
+                'document_root': os.path.join(settings.BASE_DIR, 'frontend/public'),
                 'show_indexes' : True
             }
         ),
         # Matches STATIC_URL in settings
         re_path(r'^django_static/(?P<path>.*)$', static_views.serve),
```
ssciolla, line 59: @jonespm, can you help explain why I still see `plasmic` as a directory if I go to browse the static via the browser? Maybe it's a config thing, but I haven't figured it out yet.

ssciolla:
Okay, thanks for the feedback @jonespm. I'm going to merge in an effort to keep things moving. I'm sure we'll need to revisit aspects of the UI but they should be pretty straightforward to update.

PR #162, opened 2022-05-10
Fixes #159 - Wrong URL in DRF for images when running through proxy/loophole

### LINKED ISSUE #159: Wrong URL in DRF for images when running through Loophole
When running in Loophole, images are having the URL https://127.0.0.1 instead of the original URL. It looks like the solution is to set 

USE_X_FORWARDED_HOST=True in the config

This should be fine to use for all settings, though may need to be configurable.

jonespm:


backend/settings.py
```diff
@@ -230,6 +230,8 @@
     CSRF_TRUSTED_ORIGINS = os.getenv("CSRF_TRUSTED_ORIGINS", [])
     SESSION_COOKIE_SECURE = True
     SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+    # To set the correct hostname for DRF when running https
+    USE_X_FORWARDED_HOST = True
```
ssciolla, line 234: A more general comment may be appropriate, like "# Enables proxies that set headers".

ssciolla (approved):
Not sure if it needs to be configurable or not (I'll leave that up to your determination), but it worked fine for me with `loophole` and `ngrok`.

PR #163, opened 2022-05-11
#161 created a class for canvas API LTI

### LINKED ISSUE #161: Create class or service for backend Canvas API LTI-related interactions
- [ ] Create a class that will be responsible for creating a requestor object (either `CanvasApi` or something else like `requests.session`) and making the Canvas API calls related to LTI tools that the app will need to perform.
- [ ] Add a method that can get the LTI tools available in a given course context (either using a `CanvasApi` method or the Canvas REST API directly, https://canvas.instructure.com/doc/api/external_tools.html#method.external_tools.show).

One idea of what this might look like is:
```
class CanvasLtiManager:
    requestor: CanvasApi
    course_id: int

    def __init__(self, access_token: str, course_id: int):
        # initialize requestor
        # initialize course ID
        pass

    def get_tools_available_in_course():
        # return a list of dictionaries containing some kind of subset of properties returned from Canvas for each LTI tool
        # I believe dictionary only needs to be the ID of the tool and True or False if it's enabled
	# We may add more properties later. It's going to be "something" like this call. 
        course = requestor.get_course(self.course_id)
	for tab in course.get_tabs():
            # Use the html_url to determine the tool id
            # Only return tools that have external_tools in them
	    # The last value after the / is the id
	    if "external_tools" in tab.html_url:
                print (tab.html_url)
		# I believe this can be used to tell whether it's hidden or not
		hidden = (hasattr(tab, 'hidden'))
		# The last value after the / is the id, pull that out? Or maybe just usethe tab.id? 

    # To be handled later
    def add_lti_to_nav(tab_id: str):
        pass 

    def remove_lti_from_nav(tab_id: str):
        pass
```

The add and remove may have to call this API on tabs.
https://canvas.instructure.com/doc/api/tabs.html#method.tabs.update

I'm not sure if the CanvasAPI is directly compatible, you may have to call get_tabs then update the tab now. We'll have to see if there's a way to call this directly maybe just passing the tab id. 

Resource(s):
- https://canvasapi.readthedocs.io/en/stable/course-ref.html#canvasapi.course.Course.get_external_tools
- https://canvasapi.readthedocs.io/en/stable/getting-started.html

show981111:
resolved #161 
created a class for canvas API LTI and added a unit test and command line for testing


backend/canvas_app_explorer/tests.py
```diff
@@ -1,3 +1,11 @@
...
+from .canvas_lti_manager import CanvasLtiManager
+from django.conf import settings
 # Create your tests here.
+class TestCanvasLtiManager(TestCase):
+    def test_getExTools(self):
+        API_URL = 'https://umich-dev.instructure.com'
+        API_KEY =  settings.TEST_API_KEY
+        manager = CanvasLtiManager(API_URL=API_URL, API_KEY=API_KEY, course_id=201)
+        res = manager.get_tools_available_in_course();
+        self.assertEqual(len(res),10 )
```
ssciolla, line 11:
```suggestion
        self.assertEqual(len(res), 10)

```

backend/canvas_app_explorer/tests.py
```diff
@@ -1,3 +1,11 @@
 from django.test import TestCase
-
+from .canvas_lti_manager import CanvasLtiManager
+from django.conf import settings
 # Create your tests here.
+class TestCanvasLtiManager(TestCase):
+    def test_getExTools(self):
+        API_URL = 'https://umich-dev.instructure.com'
+        API_KEY =  settings.TEST_API_KEY
+        manager = CanvasLtiManager(API_URL=API_URL, API_KEY=API_KEY, course_id=201)
```
ssciolla, line 9: Make configurable

backend/canvas_app_explorer/tests.py
```diff
@@ -1,3 +1,11 @@
...
+        res = manager.get_tools_available_in_course();
```
ssciolla, line 10:
```suggestion
        res = manager.get_tools_available_in_course()
```

backend/canvas_app_explorer/tests.py
```diff
@@ -1,3 +1,11 @@
 from django.test import TestCase
-
+from .canvas_lti_manager import CanvasLtiManager
+from django.conf import settings
 # Create your tests here.
+class TestCanvasLtiManager(TestCase):
+    def test_getExTools(self):
```
ssciolla, line 6:
```suggestion
    def test_get_ex_tools(self):
```

requirements.txt
```diff
@@ -22,3 +22,6 @@ https://github.com/Harvard-University-iCommons/django-canvas-oauth/archive/v1.1.
 
 # For serving static files
 whitenoise==5.3.0
+
+#canvasApi
+canvasapi
```
ssciolla, line 27:
```suggestion
canvasapi
```

backend/canvas_app_explorer/tests.py
```diff
@@ -1,3 +1,12 @@
...
+from django.conf import settings
 # Create your tests here.
+class TestCanvasLtiManager(TestCase):
+    def test_get_ex_tools(self):
+        API_URL =  settings.TEST_API_URL
+        API_KEY =  settings.TEST_API_KEY
+        COURSE_ID = settings.TEST_COURSE_ID
+        manager = CanvasLtiManager(API_URL=API_URL, API_KEY=API_KEY, course_id=COURSE_ID)
+        res = manager.get_tools_available_in_course()
+        self.assertEqual(len(res), 10)
```
ssciolla, line 12: I think maybe we could just check if each of the elements of `res` is an `ExternalTool`. So, using `isinstance(elem, ExternalTool)`. This won't consistently be 10.

requirements.txt
```diff
@@ -22,3 +22,5 @@ https://github.com/Harvard-University-iCommons/django-canvas-oauth/archive/v1.1.
 
 # For serving static files
 whitenoise==5.3.0
+
+canvasapi
```
jonespm, line 26: Please add a specific version to this. I think the latest is 2.2.0.

```suggestion
canvasapi==2.2.0
```

backend/canvas_app_explorer/canvas_lti_manager.py
```diff
@@ -0,0 +1,21 @@
+from typing import List
+from canvasapi import Canvas
+from .data_class import ExternalTool
+
+class CanvasLtiManager:
+    def __init__(self, API_URL: str, API_KEY: str, course_id: int):
```
jonespm, line 6: These parameters should be lowercase. There will be some other changes if these changed.

```suggestion
    def __init__(self, api_url: str, api_key: str, course_id: int):
```

jonespm (changes requested):
A few minor changes but looks pretty good for this phase! Thanks!

jonespm (approved):

backend/canvas_app_explorer/data_class.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass
+
+@dataclass
+class ExternalTool:
+    label: str
+    id: int
+    is_hidden: bool
```
ssciolla, line 7: I was thinking we may want to rename this at some point, something like `ExternalToolContext`. We already have a way of representing external tools with `models.LtiTool`, this just kinda extends it with some on information about navigation settings in a given context. I think I'm okay leaving it for now, but it may make sense soon as we try to combine this data with the `LtiTool` data.

backend/canvas_app_explorer/tests.py
```diff
@@ -1,3 +1,16 @@
...
+from .canvas_lti_manager import CanvasLtiManager
+from backend.canvas_app_explorer.data_class import ExternalTool
 
 # Create your tests here.
+class TestCanvasLtiManager(TestCase):
+    def test_get_ex_tools(self):
+        API_URL = settings.TEST_API_URL
+        API_KEY = settings.TEST_API_KEY
+        COURSE_ID = settings.TEST_COURSE_ID
+        manager = CanvasLtiManager(api_url=API_URL, api_key=API_KEY, course_id=COURSE_ID)
```
ssciolla, line 13: You could just use `settings.API_URL`, etc. in-line when instantiating `CanvasLtiManager`.

ssciolla (approved):
I'm leaving a few thoughts, some about development going forward, but I think this is okay to merge; we can improve on it as we expand this class. Thanks for the effort, @show981111! Glad you got the testing started here.

backend/canvas_app_explorer/canvas_lti_manager.py
```diff
@@ -0,0 +1,21 @@
...
+        self.requestor = Canvas(API_URL, API_KEY)
+
+    def get_tools_available_in_course(self) -> List[ExternalTool]:
+        ex_tools = []
+        course = self.requestor.get_course(self.course_id)
+        for tab in course.get_tabs():
+            if 'external_tools' in tab.html_url: #filter external tools from all tabs
+                ex_tools.append(ExternalTool(
+                    label=tab.label,
+                    id=int(tab.id.split('_')[-1]),
```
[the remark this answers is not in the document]
jonespm: Yeah, I agree and thought about this too. There's either going to be an full exception here (if they happen to change this pattern) or we'd have to do a try/catch.

This is [used in many places in Canvas ](https://github.com/instructure/canvas-lms/search?q=context_external_tool_) so seems like a big change that we'd probably be aware of if they ever did try to change it. 

Though even if they did something like removed the underscores and just made the id a int this would still work. The only way this would fail is if they changed the pattern completely which just looks unlikely.

backend/canvas_app_explorer/data_class.py
jonespm: In Canvas they call this a "Tab" 

We could call it something like `CanvasTab` or `PlatformTab` if we wanted to be more generic. I'm fine renaming it afterward. I kind of like `PlatformTab`.

backend/canvas_app_explorer/tests.py
jonespm: Yeah, these local variables aren't necessary anymore and probably should also be lower cased if they used. Just use the settings directly.

PR #165, opened 2022-05-13
Fixes #164 - Simplify local development to only need one forward

### LINKED ISSUE #164: Simplify local development to only need one forward
Talking about it, it looks like we can simplify localhost development to still have reload but only need one port forwarded. This was basically already setup, just needed webpack to write to disk instead of serving the content.

jonespm:


jonespm:
After this and #159 are committed this seems to work fine under Loophole.

Need to update:
https://github.com/tl-its-umich-edu/canvas-app-explorer/wiki/Configuring-LTI-1.3

jonespm (commented):
I locally ran this code and it works and looks good. I added the tool ID and If a tool was not in the navigation it showed the remove button.

I added some comments that we could resolve in future issues or they were just questions. 

Thanks! Almost. got the button working!

frontend/webpack.config.js
```diff
@@ -42,4 +39,9 @@ module.exports = {
     modules: ['node_modules'],
     extensions: ['.js', '.jsx']
   },
+  devServer: {
+    devMiddleware: {
+      writeToDisk: true,
+    }
+  }
```
ssciolla, line 46: Is this needed if you're using `webpack --watch`? You might be able to remove `webpack-dev-server` from `package.json` as well.

https://v4.webpack.js.org/guides/development/#using-watch-mode

deploy/supervisor_docker.conf
```diff
@@ -11,8 +11,8 @@ stderr_logfile_maxbytes=0
 startsecs=5
 startretries=2
 
-[program:npm]
-command=node server.js
+[program:frontend]
+command=/code/frontend/node_modules/.bin/webpack --config webpack.config.js --watch
```
ssciolla, line 15: It seems like we should be able to just run `webpack` without finding it in `node_modules`, but I can't get it to work. Weird.

ssciolla (changes requested):
I think we should remove the `webpack-dev-server` stuff, but otherwise this is working well. I had a couple other comments.

deploy/supervisor_docker.conf
ssciolla: Looks like you can create a Node script and use that here. So something like `npm run watch` and then add this to `package.json`:
```
  "scripts": {
    "watch": "webpack --config webpack.config.js --watch"
  },
```

jonespm: I tried a few things to get this to work, I can add this to the package and see if it runs. It might have been how I was installing node.

jonespm: If I make this switch it now says

```
> canvas-app-explorer@0.0.0 watch
> webpack --config webpack.config.js --watch

/usr/bin/env: ‘node’: Permission denied

```
I can run node and npm fine in the container.

ssciolla: Hm, okay, I'll try it.

jonespm: Well things are getting a little further. I'm getting permission errors on frontend/bundles now.

I think running it via this `npm run watch` is changing the user id from root and ends up causing permission problems if a local file system is bound as a different user id. Maybe I can move bundles one level down too out of this directory or something.

frontend/webpack.config.js
jonespm: Seems like it might not be needed.

jonespm:
Lots of fixes/updates from code review on https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/165/commits/b1851dac8ec618e9f98627be045d747f103e816f

ssciolla (approved):
LGTM, I left a couple more comments, but if you make changes, no need to re-request another review. You can also create new issues if you want.

docker-compose.yml
```diff
@@ -35,7 +35,6 @@ services:
       - ${HOME}/mylasecrets:/secrets
     ports:
       - "5000:5000"
-      - "3000:3000"
     container_name: canvas_app_explorer
```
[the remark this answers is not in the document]
jonespm: About bundles you mean? I usually just look at the logs to see when things have recompiled. We bind if if there isn't any issues.

docker-compose.yml
```diff
@@ -27,15 +27,12 @@ services:
       args:
         TZ: ${TZ}
     volumes:
-      - .:/code:delegated
-      # use the container's node_modules and bundles folder (don't override)
-      - /code/frontend/node_modules/
-      - /code/frontend/bundles/
-      - /code/static/
+      # Mount only the needed directories for local development
+      - ./frontend/components:/code/frontend/components
```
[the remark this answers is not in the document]
jonespm: Yeah, this makes sense to me to re-arrange this. It might be easier to do this on a separate PR after this one. It does reload components right now but I can see switching this over to having the reloaded code in it's own folder. 

It was either that or try to get node_modules somewhere else and that had it's own challenges.

dockerfiles/Dockerfile
```diff
@@ -17,16 +17,11 @@ RUN pip install --no-cache-dir -r requirements.txt
...
-ENV NVM_DIR=/root/.nvm
-RUN . "$NVM_DIR/nvm.sh" && nvm install ${NODE_VERSION} && \
-    . "$NVM_DIR/nvm.sh" && nvm use v${NODE_VERSION} && \
-    . "$NVM_DIR/nvm.sh" && nvm alias default v${NODE_VERSION}
-ENV PATH="/root/.nvm/versions/node/v${NODE_VERSION}/bin/:${PATH}"
-RUN node --version && npm --version
 
-COPY /frontend/package*.json /code/frontend/
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash - && \
+    apt install -y nodejs
```
ssciolla, line 22: @jonespm, I guess we need to change this in `Dockerfile.openshift` too?

jonespm: Good idea to keep it consistent. It doesn't actually use the node that's installed there at runtime, just to install the npm packages and run webpack so it probably could even remove it to save some space afterward.

ssciolla: Good point. `Dockerfile.openshift` looks like it could be optimized in a couple ways. I was looking at some of them when thinking about TypeScript.

PR #167, opened 2022-05-17
Filter tools based on course context, process navigation enabled flag (#166)

### LINKED ISSUE #166: Filter tools displayed based on availability in course context, show navigation status in UI
Now that we have a `CanvasLTIManager` class (#161), we should add `canvas_id` to the `LtiTool` model and filter the tools shown based on whether they are available in the course context. This will also involve adding `course_id` to the Django session and some modifications to the UI to process a flag on whether each tool is enabled or in the navigation.

ssciolla:
This PR aims to resolve #166. Not sure it is an optimal implementation, but it seems to work.


backend/canvas_app_explorer/serializers.py
```diff
@@ -10,7 +14,16 @@ class Meta:
...
     canvas_placement_expanded = CanvasPlacementSerializer(read_only=True, many=True, source="canvas_placement")
+    enabled = serializers.SerializerMethodField()
 
     class Meta:
         model = models.LtiTool
         fields = '__all__'
+
+    def get_enabled(self, obj: models.LtiTool) -> Union[bool, None]:
+        available_tools = self.context['available_tools']
+        matches: List[ExternalTool] = list(filter(lambda x: x.id == obj.canvas_id, available_tools))
```
jonespm, line 25: Could you leave a short code comment before this line? I'm not sure to describe what this all is doing here.

jonespm, line 25: Could you leave a short code comment before this line? I'm not sure to describe what this all is doing here.

backend/canvas_app_explorer/serializers.py
```diff
@@ -10,7 +14,16 @@ class Meta:
 class LtiToolSerializer(serializers.ModelSerializer):
     # Addiitonal expanded ead only field to make it easier on the front-end
     canvas_placement_expanded = CanvasPlacementSerializer(read_only=True, many=True, source="canvas_placement")
+    enabled = serializers.SerializerMethodField()
 
     class Meta:
         model = models.LtiTool
         fields = '__all__'
+
+    def get_enabled(self, obj: models.LtiTool) -> Union[bool, None]:
```
jonespm, line 23: Something I'd like to get better about (again) is using docstrings. I used these consistently in Java and more consistently in MyLA but I think not as much here. There's an [extension for this ](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring) but I'm not sure the format. to use. Sphinx and "one-line Sphnix" seem the most popular but Google looks easiest to read, though taking up more space. 

We can go back and do this on another issue too, or start with this one!

backend/canvas_app_explorer/views.py
```diff
@@ -1,15 +1,33 @@
...
-    queryset = models.LtiTool.objects.all()
     serializer_class = serializers.LtiToolSerializer
+    permission_classes = [permissions.DjangoModelPermissions]
+    queryset = models.LtiTool.objects.all()
+
+    def get_serializer_context(self):
+        context = super(LTIToolViewSet, self).get_serializer_context()
+        request: Request = context['request']
+        access_token = get_oauth_token(request)
+        manager = CanvasLtiManager(
```
jonespm, line 21: We probably could simplify this `CanvasLtiManager` so you only need to call it with 2 parameters most of the time. (Or all of the time)

So you'd be like `CanvasLtiManager(api_key=access_token, course_id=request.session['course_id']`). If could just pull that `api_url` parameter directly.

Actually, we could probably also just change `CanvasLtiManger` to just take a single parameter request. Then it would be even less code outside. I just didn't see how it was going to be used yet.

jonespm, line 21: We probably could simplify this `CanvasLtiManager` so you only need to call it with 2 parameters most of the time. (Or all of the time)

So you'd be like `CanvasLtiManager(api_key=access_token, course_id=request.session['course_id']`). If could just pull that `api_url` parameter directly.

Actually, we could probably also just change `CanvasLtiManger` to just take a single parameter request. Then it would be even less code outside. I just didn't see how it was going to be used yet.

backend/canvas_app_explorer/views.py
```diff
@@ -1,15 +1,33 @@
...
+            request.session['course_id']
+        )
+        context['available_tools'] = manager.get_tools_available_in_course()
+        return context
 
-    permission_classes = [permissions.DjangoModelPermissionsOrAnonReadOnly]
+    def get_queryset(self):
+        available_tool_ids = [t.id for t in self.get_serializer_context()['available_tools']]
+        return models.LtiTool.objects\
+            .filter(canvas_id__isnull=False)\
```
jonespm, line 32: I _think_ this should be written as one filter

`.filter(canvas_id__isnull=False, canvas_id__in=available_tool_id)`

jonespm, line 32: I _think_ this should be written as one filter

`.filter(canvas_id__isnull=False, canvas_id__in=available_tool_id)`

backend/canvas_app_explorer/views.py
```diff
@@ -1,15 +1,33 @@
...
     """
-    API endpoint that allows users to be viewed or edited.
+    API endpoint that lists LTI tools available in the course context
     """
-    queryset = models.LtiTool.objects.all()
     serializer_class = serializers.LtiToolSerializer
+    permission_classes = [permissions.DjangoModelPermissions]
+    queryset = models.LtiTool.objects.all()
+
+    def get_serializer_context(self):
```
jonespm, line 17: I don't think I've ever used this before, is this something with DRF? Like to pass data back?

backend/canvas_app_explorer/serializers.py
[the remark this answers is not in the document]
ssciolla: `available` is different from `enabled`, at least as I have formulated it. `available` means that the tool is available in the course context, either to be added to the navigation or removed. `enabled` means that it is in the navigation.

backend/canvas_app_explorer/serializers.py
ssciolla: I can add that, yeah. It's basically doing a search for the tool in the available tools by canvas ID.

backend/canvas_app_explorer/views.py
ssciolla: Yeah, that's fine, forgot you can do that.

backend/canvas_app_explorer/views.py
ssciolla: I had never either. It's mentioned lightly in the documentation here: https://www.django-rest-framework.org/api-guide/generic-views/

I think this SO post gave me the idea to use it: https://stackoverflow.com/a/53863873 Again, I'm still not positive this is the best way to do it.

backend/canvas_app_explorer/serializers.py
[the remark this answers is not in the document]
ssciolla: We could change this to `navigation_enabled` and `get_navigation_enabled` if we thought that was clearer.

backend/canvas_app_explorer/views.py
ssciolla: Done. See https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/167/commits/71e1c4b32b1cee0f1151d26aa020294fdf3af80a

backend/canvas_app_explorer/serializers.py
ssciolla: Done. See https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/167/commits/f0237d97daddae246615d8da71c40422eeaa8ea5

backend/canvas_app_explorer/serializers.py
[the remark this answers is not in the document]
ssciolla: I made this change, seems like the extra info may help readability.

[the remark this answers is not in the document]
ssciolla: I added some basic docstrings with descriptions for the methods I added in the backend. I like having these, but I'd rather not decide on a format or extension right now. Feel free to create a new issue if you'd like.

backend/canvas_app_explorer/serializers.py
```diff
@@ -10,7 +14,17 @@ class Meta:
...
+        available_tools = self.context['available_tools']
+        # Search in tools available in the context for a canvas ID matching the model instance
+        matches: List[ExternalTool] = list(filter(lambda x: x.id == obj.canvas_id, available_tools))
+        if len(matches) > 0:
+            first_match = matches[0] # Canvas IDs should be unique
+            return not first_match.is_hidden
+        return None
```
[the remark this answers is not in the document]
ssciolla: I did this. It seemed like it'd surface the bug faster if these edge cases occur. See https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/167/commits/bc22e72852a5c3cafb270b8169c707f95228d01c

ssciolla:
Thanks for the review, @jonespm. If you could take a look at my changes, and respond to my comments, and give another approval, I'd appreciate it.

backend/canvas_app_explorer/views.py
jonespm: Yeah,  I was thinking that just to simplify all of the code that will be calling into it.

I mean we could make the method take multiple parameters with some default values and figure out what to do based on what was passed too since you can't overload constructors in Python.

I think this is fine for now but I don't really want to copy/paste 3-4 LOC every time we initialize this object. Maybe I'll revise this when I add more methods to the API.

backend/canvas_app_explorer/serializers.py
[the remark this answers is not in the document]
jonespm: Okay, I'll create a new issue. Thanks!

jonespm (approved):

backend/canvas_app_explorer/views.py
ssciolla: I get it, you don't want to repeat things too much. However, I also realized that making the manager class know about requests and environment variables limits our ability to write tests, I think.

If we wanted to make both of us happy, we could create a factory?

Something like this:
```
class DjangoCourseLTIManagerFactory():
    
    def __init__(api_url: string):
        self.api_url = api_url

    createManager(request: Request) -> CanvasLTIManager:
        # Logic setting up access_token and course ID go here
        course_id = request.session['course_id']
        token = get_access_token(request)
        return CanvasLTIManager(self.api_url, token, course_id)
```

And then initialize a factory instance somewhere and share it between modules. WDYT?

ssciolla: I'll merge this PR, and we can discuss this idea as part of your work.

backend/canvas_app_explorer/views.py
```diff
@@ -1,15 +1,39 @@
...
+        """
+        Extends context for LtiToolSerializer to include data (from the Cavnas API) about tools available
```
ssciolla, line 19:
```suggestion
        Extends context for LtiToolSerializer to include data (from the Canvas API) about tools available
```

backend/canvas_app_explorer/views.py
```diff
@@ -1,15 +1,39 @@
...
+        in the course and whether they are enabled for the navigation.
```
ssciolla, line 20:
```suggestion
        Extends context for LtiToolSerializer to include data (from the Canvas API) about tools available
        in the course and whether they are enabled for the navigation
```

frontend/components/ToolCard.jsx
```diff
@@ -8,6 +7,7 @@ import {
 import { styled } from '@mui/material/styles';
 
 import DataElement from './DataElement';
+import AddRemoveButton from './AddRemoveButton';
```
ssciolla, line 10:
```suggestion
import AddRemoveButton from './AddRemoveButton';
import DataElement from './DataElement';

```

frontend/components/ToolCard.jsx
```diff
@@ -1,14 +1,15 @@
...
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import {
   Button, Card, CardActions, CardContent, CardMedia, Collapse, Grid, IconButton, Typography
 } from '@mui/material';
 import { styled } from '@mui/material/styles';
 
+import AddRemoveButton from './AddRemoveButton';
 import DataElement from './DataElement';
 
+
```
ssciolla, line 12:
```suggestion
```

backend/canvas_app_explorer/views.py
jonespm: This is an idea. I think we'd probably want two methods, one that can take a Request, or one that can take course id and token. 

And also api_url could be defaulted `settings.CANVAS_OAUTH_CANVAS_DOMAIN` if not set. 

Fine to work on this independently, and maybe don't even need it now. Just want to make this as easy/brief to call if we'll be calling it a lot.

Issue #168, opened 2022-05-18
Clean up console.log in the frontend Home.jsx


jonespm:
I think it's fine to have console log messages in the frontend but not sure how helpful the message like

"WE SHOULD GO IN HERE, show all data"
is when it's displayed. Either removed these logs or improve them.

https://github.com/tl-its-umich-edu/canvas-app-explorer/blob/354b35d21c82ec3f96d932918aaf0935c21ed5ba/frontend/components/Home.jsx#L43

Issue #137, opened 2022-04-16
Consider switching to TypeScript


ssciolla:
There isn't a lot of custom code in the frontend right now, so we might want to try regenerating the Plasmic files as TypeScript and use that going forward. I have liked the stronger typing and extra language features, and it's especially nice to be able to write interfaces for custom components and get good linting feedback.

PR #171, opened 2022-05-25
Switch to TypeScript (#136, #137, #168)

### LINKED ISSUE #136: Set up frontend linting approach
I noticed when poking around in the `frontend` that there doesn't seem to be a linting approach adopted for JavaScript yet. I guess my recent experience has been with `eslint`, but we could look into other options. I've liked using `standard`-like rules, but the tooling seems to be lagging a little bit on updates.

ssciolla:
The PR aims to resolve #136, #137, and #168.


dockerfiles/Dockerfile.openshift
```diff
@@ -1,43 +1,42 @@
...
+WORKDIR /build/
+
+COPY frontend .
+RUN npm install
+
+RUN npm run build:frontend
+
+# main stage
+
+FROM docker-registry.default.svc:5000/openshift/python:3.8-slim AS main
```
ssciolla, line 13: I didn't test pulling from the OpenShift registry, I just removed the namespace to pull from Docker Hub. Seems to work though, should be fine with deployments.

frontend/.eslintrc.js
```diff
@@ -0,0 +1,47 @@
...
+      'unix'
+    ],
+    'quotes': [
+      'error',
+      'single'
+    ],
+    'semi': [
+      'error',
+      'always'
+    ]
```
ssciolla, line 45: Not sure about enforcing this, but we were kinda using them sporadically, I guess I figured we may as well be consistent.

I'm hesistant to adopt `standard` right now because the `standard` + ESLint + TypeScript + React stuff is a little behind. I figured we could just define our own rule set as we go for now.

frontend/components/Home.tsx
```diff
@@ -0,0 +1,78 @@
...
+      const data: Tool[] = await response.json();
+      // sort data alphabetically by name
+      data.sort((a, b) => (a.name.toLowerCase() > b.name.toLowerCase())
+        ? 1
+        : ((b.name.toLowerCase() > a.name.toLowerCase()) ? -1 : 0)
+      );
+      setTools(data);
+    };
+    fetchToolData();
+  }, []);
```
ssciolla, line 44: This was tidied up in the process of converting; we didn't need a separate `useEffect` or state variable for handling the filter. The component will re-render and filter the tools down when `searchFilter` changes. See line 50.

ssciolla: I think I resolved #168 here. I'm mostly of the opinion that `console.log` should be used sparingly or not at all, but `console.error` is fine to use when needed. I use `console.log` a lot in debugging though.

frontend/components/ToolCard.tsx
```diff
@@ -2,35 +2,60 @@ import React, { useState } from 'react';
...
+  }
+
+  let mainImageProps: CardImageProps = {
+    component: 'img',
+    height: 150,
+    alt: tool.main_image_alt_text ?? 'Image of tool in use'
+  };
+  if (tool.main_image !== null) {
+    mainImageProps = { image: tool.main_image, ...mainImageProps };
+  }
```
ssciolla, line 50: I had to do some strange stuff here to placate the conditional interface on `CardMedia` and also handle the `null` cases. I may return to to this later, especially if we add a default image, but this should be functional for now.

frontend/package.json
```diff
@@ -17,14 +17,21 @@
     "react-dom": "^17.0.2"
   },
   "devDependencies": {
-    "@babel/core": "^7.9.6",
-    "@babel/preset-env": "^7.9.6",
-    "@babel/preset-react": "^7.9.4",
-    "babel-loader": "^8.1.0",
```
ssciolla, line 23: As with CCM, I'm just relying on `tsc` (the TypeScript compiler) to do the work that we used to use Babel to do. See https://www.typescriptlang.org/docs/handbook/babel-with-typescript.html for some additional info.

ssciolla:
@jonespm, we may surface some issues in the review, particularly around Docker changes, but I think it's ready for you to look at. Thanks

frontend/components/AddRemoveButton.tsx
```diff
@@ -1,12 +1,23 @@
 import React from 'react';
 import AddIcon from '@mui/icons-material/Add';
 import RemoveIcon from '@mui/icons-material/Remove';
-import { Button, SvgIcon, Typography } from '@mui/material';
+import { Button, SvgIconProps, Typography } from '@mui/material';
 
-export default function AddRemoveButton (props) {
+interface AddRemoveButtonProps {
```
jonespm, line 6: Have you considered extending Button like described here? https://mui.com/material-ui/guides/typescript/#complications-with-the-component-prop

Is there an reason to do a custom interface? I'm just learning here so just tying to find best practice for this.

frontend/interfaces.ts
```diff
@@ -0,0 +1,21 @@
+interface CanvasPlacement {
+  name: string
+}
+
+interface Tool {
+  id: number,
+  name: string,
+  canvas_id: number,
+  logo_image: string | null,
```
jonespm, line 9: All of these This might look better with a user-defined type like 

`type Nullable<T> = T | null;`

I suppose the might be optional but probably easier for the API to just always return something. Though does it actually return a null now or just an empty string? I'd have to check unless you already have.

jonespm (approved):
A few comments but this builds and runs and looks like a good improvement! Thanks!

frontend/components/Home.tsx
ssciolla: Yeah, maybe there are some FE strategies for logging that I haven't explored.

frontend/components/AddRemoveButton.tsx
ssciolla: Maybe there's something I can simplify around `ButtonProps`, I will take a look. We could have some of these interfaces extend from `mui.ButtonProps`, but we are setting some of these values internally based on the `navEnabled` flag.

Extending `Button`, I'm not sure about. I think I'm kind of following the composition approach they describe, although I'm not really using the `component` prop (`button` is fine for our purpose), so this doc is only somewhat relevant.

frontend/components/Header.tsx
```diff
@@ -2,7 +2,11 @@ import React from 'react';
 import { Grid, Paper, TextField, Typography } from '@mui/material';
 import SearchIcon from '@mui/icons-material/Search';
 
-export default function Header (props) {
+interface HeaderProps {
```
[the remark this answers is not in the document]
ssciolla: I'm not opposed to switching to an `AppBar` implementation, it would need some customization. I'm not sure about the profile icon, we're not yet sending user info to the FE. I view this as out of scope for this issue, right?

frontend/interfaces.ts
ssciolla: I'm pretty sure it just returns `null` for now, and I believe I checked. The `Nullable<T>` is interesting, though I think it's basically just as long. I think I'll keep it this way for now, unless you object.

frontend/components/AddRemoveButton.tsx
ssciolla: @jonespm, I thought about this some more, and I think you were on to something. The more I looked at this, the more I wasn't crazy about the branching styling, and realized we could rely more on `ButtonProps`. I refactored it here a bit: https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/171/commits/015fc5a2285049a4fcb2452ef897e009019613d9

frontend/components/Header.tsx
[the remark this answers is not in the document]
ssciolla: I created a separate issue, #182.

frontend/components/AddRemoveButton.tsx
jonespm: Ah I see, yeah looks good! I think it's easier to read than the other format.

dockerfiles/Dockerfile.openshift
jonespm: When I started this last year I wanted to see if there was a better way to avoid multi-stage builds, but as the year went on and we looked to finally be moving off Openshift 3, as well as Github Container Registry, it having mutlstage just was less of a problem. I think we can actually consolidate these files once we get to ROSA.

PR #173, opened 2022-05-25
Rework `LtiToolViewSet` to prevent multiple API calls (#172)

### LINKED ISSUE #172: Canvas API gettting called multiple times via `get_serializer_context`
I noticed after #167 was merged that the Canvas API was getting called multiple times (3?) to fetch tool navigation information. This seems to be because of `get_serializer_context` works (I also used it twice directly).

https://github.com/tl-its-umich-edu/canvas-app-explorer/blob/f9cdee72e13ca57b159ae4f904765caa1a716714/backend/canvas_app_explorer/views.py#L17-L39

Looks like I can circumvent this issue by using a more generic view type, `viewsets.ViewSet`, and passing the extra data to the serializer class via a context parameter.

ssciolla:
The PR aims to resolve #172.

### DIFF
--- a/backend/canvas_app_explorer/views.py
+++ b/backend/canvas_app_explorer/views.py
@@ -1,39 +1,35 @@
+import logging
+
 from canvas_oauth.oauth import get_oauth_token
 from django.conf import settings
 from rest_framework import permissions, viewsets
 from rest_framework.request import Request
+from rest_framework.response import Response
 
 from backend.canvas_app_explorer import models, serializers
 from .canvas_lti_manager import CanvasLtiManager
 
-class LTIToolViewSet(viewsets.ModelViewSet):
+logger = logging.getLogger(__name__)
+
+class LTIToolViewSet(viewsets.ViewSet):
     """
     API endpoint that lists LTI tools available in the course context
     """
-    serializer_class = serializers.LtiToolSerializer
-    permission_classes = [permissions.DjangoModelPermissions]
-    queryset = models.LtiTool.objects.all()
+    permission_classes = [permissions.IsAuthenticated]
 
-    def get_serializer_context(self):
-        """
-        Extends context for LtiToolSerializer to include data (from the Canvas API) about tools available
-        in the course and whether they are enabled for the navigation
-        """
-        context = super(LTIToolViewSet, self).get_serializer_context()
-        request: Request = context['request']
+    def list(self, request: Request):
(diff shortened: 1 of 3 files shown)


backend/canvas_app_explorer/views.py
```diff
@@ -1,39 +1,35 @@
...
+logger = logging.getLogger(__name__)
+
+class LTIToolViewSet(viewsets.ViewSet):
     """
     API endpoint that lists LTI tools available in the course context
     """
-    serializer_class = serializers.LtiToolSerializer
-    permission_classes = [permissions.DjangoModelPermissions]
-    queryset = models.LtiTool.objects.all()
+    permission_classes = [permissions.IsAuthenticated]
```
jonespm, line 18: If changing this it may need to be [IsAuthenticatedOrReadOnly](https://www.django-rest-framework.org/api-guide/permissions/#isauthenticatedorreadonly).

I believe that this view was not intended to be protected and we'd eventually want to have it available to the public #27. I think this is requiring authentication now after this change from DjangoModelPermissions.

jonespm (approved):
Thanks I never noticed this. I guess I need to read some more about how DRF works with these.

backend/canvas_app_explorer/views.py
ssciolla: So, wouldn't instructors also be authenticated (i.e. successfully launch the tool)?

ssciolla: We need the launch to work here, or else where is no course ID.

jonespm: Hmm, maybe we'll have to table this for now and work on it as part of #27 if we do that. 

The idea for that was it would show the list but not have the enable/disable information so it wouldn't need the course ID. It might just need a separate class an a little refactor to have 2 different API's and views depending on if the user was authenticated or not. 

The idea of #27 was that this could eventually replace a page like https://documentation.its.umich.edu/node/1222 but look nicer, be easier to update and have more information. So completely public facing.

ssciolla: Hm, yeah, we could probably figure that out with a separate API endpoint/view, branching behavior, and/or with a modified serializer. As it is now, this wouldn't work with a completely public endpoint; we'll have to work on that separately.

ssciolla:
Thanks, @jonespm, I'm still learning as well.

Issue #186, opened 2022-06-06
MySQL 5.7 problem with MacBook M1


zqian:
Docker-compose up throws the following error with MySQL 5.7 on MacBook M1
```
➜  canvas-app-explorer_copy git:(main) ✗ docker-compose up
WARNING: The TZ variable is not set. Defaulting to a blank string.
Creating network "canvas-app-explorer_copy_default" with the default driver
Pulling mysql (mysql:5.7)...
5.7: Pulling from library/mysql
ERROR: no matching manifest for linux/arm64/v8 in the manifest list entries
```

We've upgrade the dev MySQL server to MySQL v8.0.27, so it makes sense to update the docker-compose file to use MySQL 8

jonespm:
After switching to MySQL 8 my local startup is taking nearly 3 minutes on a empty database. Adding the 2 options to the command in docker-compose.yml pushes it down to like 1 second.  Probably will have to put these in but I'm not sure if there's any downsides.

Saw this mentioned on https://stackoverflow.com/a/42221456/3708872

`'--skip-name-resolve', '--skip-external-locking'`

PR #187, opened 2022-06-06
issue #186 update to MySQL v8

zqian:
- [x] update MySQL version to v8

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,13 +1,12 @@
 version: "3.9"
    
 services:
-
   mysql:
-    image: mysql:5.7
-    restart: unless-stopped
-    command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci']
-    entrypoint: ['docker-entrypoint.sh', '--default-authentication-plugin=mysql_native_password']
+    image: mysql/mysql-server:8.0
+    restart: on-failure
+    command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci', '--socket=/tmp/mysql.sock']
     environment:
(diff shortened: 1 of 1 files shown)


ssciolla (approved):
This seems to work fine, I didn't even have to rebuild the `.data` volume. @jonespm, can you also review? Seems like you've probably discussed/approved a lot of these settings changes for other development work.

docker-compose.yml
```diff
@@ -17,8 +16,7 @@ services:
     ports:
       - "6306:3306"
     volumes:
-      - ./.data/mysql:/var/lib/mysql:delegated
-      - ./mysql:/docker-entrypoint-initdb.d:ro
```
jonespm, line 21: I think we need to leave this in here otherwise the mysql/init.sql won't run on initial startup. This caused an issue for Yong recently. I'm not sure why we didn't need that for our other databases.

jonespm: It seems to be working fine for me without this, but it's taking a long time to startup and creating almost 200MB of large files. Though looking back at 5.7 it also created large ib_logfile0 and ib_logfile1, but no other large files. so took up about half the space. Maybe it's not worthwhile to fix here.

I'd `kind of` like to get all of our T&L Docker databases to use the same image and name. This would make it a little easier to work with locally as the number of apps we support grow.

I guess this is a different issue though.

jonespm (approved):

docker-compose.yml
zqian: Thanks for reviewing. I will commit the PR for now.

jonespm:
I pushed the previous fix to run the init script. It can connect now but I don't have it configured.

Do we have the way to run this script and the environment variables documented anywhere?

Issue #188, opened 2022-06-06
Local error about missing webpack types


jonespm:
Seeing this error when I start up locally, not seeing it in Openshift. 

I ran that `npm i --save-dev @types/webpack` and that seemed to fix it.

Have you seen this @ssciolla ?

```
TSError: ⨯ Unable to compile TypeScript:
webpack/webpack.dev.ts(1,26): error TS7016: Could not find a declaration file for module 'webpack'. '/code/frontend/node_modules/webpack/lib/webpack.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/webpack` if it exists or add a new declaration (.d.ts) file containing `declare module 'webpack';`
```

PR #190, opened 2022-06-06
Correct and pare down `.dockerignore` (#189)

### LINKED ISSUE #189: Fix `.dockerignore`
Right now, `.dockerignore` seems to mostly be a `.gitignore` template. This is an issue because the patterns aren't the same for `.dockerignore` (and there may be cases where we want something in the Docker image, but not in `git`). It's especially an issue right now because `node_modules` are not being excluded, so `node_modules` installed locally will get copied in and wipe out anything that was generated during `docker compose build`.

ssciolla:
This PR aims to resolve #189 (and maybe #188).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,136 +1,17 @@
-# Our project specific
-saml
-bq_cred.json
+# Secrets
+**/*.env*
+
+# Node dev
+**/node_modules/
+
+# Python dev
+venv/
+**/__pycache__/
+.mypy_cache/
+
+# Database
+.data
+
+# IDE/OS stuff
 .idea/
 .DS_Store
-*.sql
-*.md
-*.sample
-*.tgz
-*.txt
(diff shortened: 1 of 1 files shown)


jonespm:
This looks to fix the problem on another issue.

I'm looking through the differences and what's on the image and also looks like it catches everything

I think we could also exclude

`**/.git`

as we aren't using it here. It is used in the MyLA start script to figure out the version but there's probably a different way to do that if we wanted that in this project than retaining all that data.

ssciolla:
> I think we could also exclude
> 
> `**/.git`

Done. I added a few more things that didn't seem necessary for the image. See https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/190/commits/b2ae64634c4f02d711ce325d4f812484ec55acc9

jonespm (approved):

PR #192, opened 2022-06-07
Fixes #179 - Don't consider HTML formatting in character count limits

### LINKED ISSUE #179: Don't consider HTML formatting in character count limits
If someone includes a link or something like that in the short description (or other fields that have character limits) it's including that as part of the character count and preventing it from saving. It should have a way to consider the max_length without any formatting characters. 

I think we'll have to use a custom max_length validator https://docs.djangoproject.com/en/4.0/ref/validators/

And maybe run this through https://docs.djangoproject.com/en/4.0/ref/utils/#django.utils.html.strip_tags.

jonespm:
I changed this class name about 10 times and tested a few different ways. Looks to be the validator pattern for accepting parameters.

### DIFF
--- a/backend/canvas_app_explorer/models.py
+++ b/backend/canvas_app_explorer/models.py
@@ -21,7 +21,7 @@ class MaxLenIgnoreHTMLValidator:
         value = strip_tags(value.strip())
         if len(value) > self.max_length:
             raise ValidationError(
-                (f'This field without tags is {len(value)}, greater than max length of {self.max_length}'),
+                (f'This field without tags is {len(value)}, which is greater than the max length of {self.max_length}.'),
             )
 
 ### Models are below
--- a/backend/canvas_app_explorer/models.py
+++ b/backend/canvas_app_explorer/models.py
@@ -1,4 +1,3 @@
-from django.core.exceptions import ValidationError
 from django.core.validators import MaxLengthValidator
 from django.db import models
 from django.utils.deconstruct import deconstructible
--- a/backend/canvas_app_explorer/models.py
+++ b/backend/canvas_app_explorer/models.py
@@ -1,9 +1,30 @@
+from django.core.exceptions import ValidationError
 from django.db import models
+from django.utils.deconstruct import deconstructible
+from django.utils.html import strip_tags
 from db_file_storage.model_utils import delete_file, delete_file_if_needed
 from tinymce.models import HTMLField
 
-# Create your models here.
+# Based on the built in validators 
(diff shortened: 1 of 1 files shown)


backend/canvas_app_explorer/models.py
```diff
@@ -1,9 +1,30 @@
...
 from db_file_storage.model_utils import delete_file, delete_file_if_needed
 from tinymce.models import HTMLField
 
-# Create your models here.
+# Based on the built in validators 
+# https://github.com/django/django/blob/main/django/core/validators.py
+# Use in your model as validators=[MaxLenIgnoreHTMLValidator(max_length=120)]
+@deconstructible
+class MaxLenIgnoreHTMLValidator:
+    max_length = 120
```
zqian, line 13: Works like magic! However, can we consolidate the max_length value definition as a constant? Looks like it is also in the database table definition.

jonespm: Yeah, this is just a default value if one isn't passed in via a parameter. The one used in the model is used. 

I was thinking afterward of just raising an exception if it wasn't set or maybe just extending the [MaxLengthValidator](https://github.com/django/django/blob/49b470b9187b6be60e573fed08c8f4a87f133750/django/core/validators.py#L433) that's there.

backend/canvas_app_explorer/models.py
```diff
@@ -1,9 +1,30 @@
...
+from django.utils.html import strip_tags
 from db_file_storage.model_utils import delete_file, delete_file_if_needed
 from tinymce.models import HTMLField
 
-# Create your models here.
+# Based on the built in validators 
+# https://github.com/django/django/blob/main/django/core/validators.py
+# Use in your model as validators=[MaxLenIgnoreHTMLValidator(max_length=120)]
+@deconstructible
+class MaxLenIgnoreHTMLValidator:
```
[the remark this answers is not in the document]
ssciolla: Maybe you could simplify here by extending `BaseValidator` like `MaxLengthValidator` does?
https://github.com/django/django/blob/49b470b9187b6be60e573fed08c8f4a87f133750/django/core/validators.py#L432-L447

ssciolla (approved):
This seems to work, so I'm approving. I had some refactoring/text change suggestions, do with them as you see fit.

backend/canvas_app_explorer/models.py
jonespm: This is outdated with the current code and no longer used.

ssciolla (approved):

PR #193, opened 2022-06-07
Fixes #110 - Make links open in new window by default

### LINKED ISSUE #110: Make links open in new window by default
When add a link in the editor, the default is current window. This doesn't add any target attribute to the anchor. 

We could see if there's an option to select "New Window" by default as we want most of the links to default to a new window.

If not maybe we can add some [javascript to just open all links from this tool in a new window.](https://stackoverflow.com/questions/12235585/is-there-a-way-to-open-all-a-href-links-on-a-page-in-new-windows) I think this would only be set if there isn't already a `target` attribute. We could also do this on the save method on the model of the backend, possibly with BeautifulSoup? 

If there's no target attribute set it should add in `target="_blank"`

jonespm:


backend/settings.py
```diff
@@ -210,9 +210,10 @@
 
 TINYMCE_DEFAULT_CONFIG = {
         'menubar': True,
-        "plugins": "advlist,autolink,lists,link,anchor,insertdatetime,media,table,paste,",
+        "plugins": "advlist,autolink,lists,link,anchor,insertdatetime,media,table,paste,wordcount",
```
ssciolla, line 213: Is there a `charcount`? Seems like that would be more useful in the `short_description` case, but maybe words are better in other fields.

jonespm: If you click the word count in the status bar it changes to character count. I didn't see any option to switch the defaults other than to use javascript which the django-tinymce plugin doesn't support yet as part of the init and it looks clunky.

They wanted me to do contribute a patch to support this to also allow for dark mode but I haven't gotten that yet. You're right I'm also not sure how it would interact with this custom validator and if it would return the same HTML-less count. I can also check that and remove it if it doesn't.

jonespm: Yeah this count looks consistent with what's displayed.

PR #195, opened 2022-06-08
Move frontend application code into `app` (#170)

### LINKED ISSUE #170: Revise the frontend layout to more efficiently bind the volume
This probably solves many use cases, but it's missing a few files, like `theme.js` and `Home.css` that might also be changed. I think we will want to just change the `frontend` project structure a bit. I'd like to see something like this:
```
frontend/
    app/
        components/
            Home.jsx
            ToolCard.jsx
            # other components go here
        index.jsx
        Home.css
        theme.js
    bundles/
    package.json
    package-lock.json
    node_modules/
    webpack.config.js
```

Then you can just volume mount `app`. This can be done now or later.

_Originally posted by @ssciolla in https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/165#discussion_r880590250_

ssciolla:
This PR aims to resolve #170.


frontend/app/index.tsx
```diff
@@ -2,9 +2,8 @@ import React from 'react';
...
 import { ThemeProvider } from '@mui/material';
 
-import theme from '../theme';
-
-import Home from './Home';
+import Home from './components/Home';
+import theme from './theme';
 
 ReactDOM.render(
   (
```
ssciolla, line 9: Note also that this file was also moved out of `components` into `app`, since it's not exporting any components.

jonespm (approved):

Issue #191, opened 2022-06-07
Wire up add and remove buttons in UI to call navigation update API


ssciolla:
Once #180 is merged, we can add API calls to the backend to use the update navigation endpoint. This work will also likely involve #181.

PR #196, opened 2022-06-09
Call API with add and remove buttons, refine loading and error strategies (#181, 191)

### LINKED ISSUE #181: Create standardized loading component or pattern
Right now we're just using a string like "Loading ..." to indicating when the UI is waiting for the results of an API call or action. Material UI has nicer waiting-related components. In coordination with the UX and accessibility teams, we will want to come up with a more noticeable/stylish solution that can be used consistently for all or most actions.

ssciolla:
This PR aims to resolve #181 and #191.

More details
- Separated API calling logic into an `api` module.
- Added `js-cookie` and process for adding a CSRF token to headers.
- Adopted `react-query` for handling asynchronous-related state (`useQuery`, `useMutation`).
- Improved loading UI appearance and accessibility (#181).
- Added API error handling and display (`ErrorsDisplay`).
- Display dismiss-able alert about refreshing the page when navigation changes are first made.
- Added `eslint` plugin for `react-hooks`.

Resources
- https://github.com/TanStack/query
- https://docs.djangoproject.com/en/4.0/ref/csrf/#ajax
- https://reactjs.org/docs/hooks-rules.html
- https://mui.com/material-ui/getting-started/overview/

### DIFF
--- a/frontend/app/api.ts
+++ b/frontend/app/api.ts
@@ -39,9 +39,15 @@ async function getTools (): Promise<Tool[]> {
   return data;
 }
 
-async function updateToolNav (canvas_tool_id: number, navEnabled: boolean): Promise<void> {
+interface UpdateToolNavData {
+  canvasToolId: number
+  navEnabled: boolean
+}
+
+async function updateToolNav (data: UpdateToolNavData): Promise<void> {
+  const { canvasToolId, navEnabled } = data;
   const body = { navigation_enabled: navEnabled };
-  const url = `${API_BASE}/lti_tools/${canvas_tool_id}/`;
+  const url = `${API_BASE}/lti_tools/${canvasToolId}/`;
   const requestInit: RequestInit = {
     method: 'PUT',
     body: JSON.stringify(body),
--- a/frontend/app/api.ts
+++ b/frontend/app/api.ts
@@ -23,7 +23,7 @@ const createErrorMessage = async (res: Response): Promise<string> => {
   }
   return (
     'Error occurred! ' +
-    `Status: ${res.status}${res.statusText !== '' ? ' ' + res.statusText : ''}` +
+    `Status: ${res.status}` + (res.statusText !== '' ? ` (${res.statusText})` : '') +
     (errorBody !== undefined ? '; Body: ' + JSON.stringify(errorBody) : '.')
   );
 };
@@ -51,7 +51,10 @@ async function updateToolNav (canvas_tool_id: number, navEnabled: boolean): Prom
     }
   };
   const res = await fetch(url, requestInit);
-  if (!res.ok) throw new Error(await createErrorMessage(res));
+  if (!res.ok) {
+    console.error(res);
+    throw new Error(await createErrorMessage(res));
+  }
   return;
 }
 
--- a/frontend/app/api.ts
+++ b/frontend/app/api.ts
@@ -53,7 +53,7 @@ async function updateToolNav (data: UpdateToolNavData): Promise<void> {
     body: JSON.stringify(body),
     headers: {
       ...BASE_MUTATION_HEADERS,
-      'X-CSRFTOKEN': getCSRFToken() ?? '' // Could throw error?
+      'X-CSRFTOKEN': getCSRFToken() ?? ''
     }
   };
   const res = await fetch(url, requestInit);
--- a/frontend/app/api.ts
+++ b/frontend/app/api.ts
@@ -1,0 +1,31 @@
+import { Tool } from './interfaces';
+
+const API_BASE = '/api';
+
+const createErrorMessage = async (res: Response): Promise<string> => {
+  let errorBody;
+  try {
+    errorBody = await res.json();
+  } catch {
+    console.error('Error body was not JSON.');
+    errorBody = undefined;
+  }
+  return (
+    'Error occurred! ' +
+    `Status: ${res.status}${res.statusText !== '' ? ' ' + res.statusText : ''}` +
+    (errorBody !== undefined ? '; Body: ' + JSON.stringify(errorBody) : '.')
+  );
+};
+
+async function getTools (): Promise<Tool[]> {
+  const url = `${API_BASE}/lti_tools/`;
+  const res = await fetch(url);
+  if (!res.ok) {
+    console.error(res);
+    throw new Error(await createErrorMessage(res));
+  }
+  const data: Tool[] = await res.json();
+  return data;
+}
+
+export { getTools };
--- a/frontend/app/api.ts
+++ b/frontend/app/api.ts
@@ -1,6 +1,17 @@
+import Cookies from 'js-cookie';
+
 import { Tool } from './interfaces';
 
+export const getCSRFToken = (): string | undefined => Cookies.get('csrftoken');
+
 const API_BASE = '/api';
+const JSON_MIME_TYPE = 'application/json';
+
+const BASE_MUTATION_HEADERS: HeadersInit = {
+  Accept: JSON_MIME_TYPE,
+  'Content-Type': JSON_MIME_TYPE,
+  'X-Requested-With': 'XMLHttpRequest'
+};
 
 const createErrorMessage = async (res: Response): Promise<string> => {
   let errorBody;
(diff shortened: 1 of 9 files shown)


frontend/app/api.ts
```diff
@@ -0,0 +1,67 @@
...
+async function updateToolNav (data: UpdateToolNavData): Promise<void> {
+  const { canvasToolId, navEnabled } = data;
+  const body = { navigation_enabled: navEnabled };
+  const url = `${API_BASE}/lti_tools/${canvasToolId}/`;
+  const requestInit: RequestInit = {
+    method: 'PUT',
+    body: JSON.stringify(body),
+    headers: {
+      ...BASE_MUTATION_HEADERS,
+      'X-CSRFTOKEN': getCSRFToken() ?? ''
```
ssciolla, line 56: We could maybe throw an error in `getCSRFToken` instead, but it seems unlikely that the CSRF token won't be there.

frontend/app/components/Home.tsx
```diff
@@ -25,34 +28,41 @@ const filterTools = (tools: Tool[], filter: string): Tool[] => {
 };
 
 function Home () {
-  const [tools, setTools] = useState<null | Tool[]>(null);
+  const [tools, setTools] = useState<undefined | Tool[]>(undefined);
+
+  const { isLoading: getToolsLoading, error: getToolsError } = useQuery('getTools', api.getTools, {
+    onSuccess: (data) => setTools(data)
+  });
```
ssciolla, line 35: In past React development, I've used a custom hook to manage loading and error state, e.g. `usePromise` in ROHQ and CCM. The TypeScript support though was always a little lacking, and I kinda figured this was probably a solved problem. I spent a significant chunk of time last week experimenting with different libraries, including `react-async` and react-async-hook`. The APIs were pretty similar, but I finally settled on [`react-query`](https://github.com/TanStack/query) since it had everything I wanted (good types, `onSuccess` hooks, large active user base, and other nice fetching features). It might be a little heavier than we need, but I think the advantages outweigh that.

frontend/app/index.tsx
```diff
@@ -1,15 +1,29 @@
...
 
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      retry: false,
+      retryOnMount: false,
+      staleTime: Infinity
+    },
+    mutations: { retry: false }
+  }
```
ssciolla, line 17: `react-query` includes a lot of retry and refetch behavior by default. We may want some of this eventually, but for now I thought it seemed unnecessary and easier for testing to disable it all. See https://react-query.tanstack.com/guides/important-defaults

frontend/app/components/Home.tsx
```diff
@@ -25,34 +28,44 @@ const filterTools = (tools: Tool[], filter: string): Tool[] => {
...
+
+  const onToolUpdate = (newTool: Tool) => {
+    /*
+    Creates new array with newTool replacing its previous version;
+    Uses function inside setState hook to handle overlapping requests
+    */
+    setTools((oldTools) => {
+      if (oldTools === undefined) throw Error('Expected tools variable to be defined!');
```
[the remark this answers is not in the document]
ssciolla: So, this isn't an API error. `createErrorMessage` is intended to process a `Response` object and return a string summarizing what happened with an API error.

I don't ever expect this to happen; this function should only be called inside of a `ToolCard`, which would need to have tool data for it to work. However, because `tools` in the `useState` hook starts off as `undefined`, it is possible if the function is used incorrectly. To satisfy TypeScript, I am throwing an error here. If you think of a better solution, please let me know.

frontend/app/api.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+  };
+  const res = await fetch(url, requestInit);
+  if (!res.ok) {
+    console.error(res);
+    throw new Error(await createErrorMessage(res));
```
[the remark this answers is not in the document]
ssciolla: Yes, I was either lazy when I set that up, or figured we'd come up with a more robust Canvas error processing strategy later on. I consider this a separate issue; I'll create one based on your comment.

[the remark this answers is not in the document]
ssciolla: See #205, thanks!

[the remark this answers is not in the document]
zqian: Thanks for tracking it.

zqian (approved):

frontend/app/api.ts
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    'Error occurred! ' +
+    `Status: ${res.status}` + (res.statusText !== '' ? ` (${res.statusText})` : '') +
+    (errorBody !== undefined ? '; Body: ' + JSON.stringify(errorBody) : '.')
+  );
+};
+
+async function getTools (): Promise<Tool[]> {
+  const url = `${API_BASE}/lti_tools/`;
+  const res = await fetch(url);
```
jonespm, line 33: Perhaps not for this issue but have you tried axios or any other libraries for doing HTTP requests? That looks like it simplifies a lot of the code repetition and has cleaner patterns that are common just doing this with fetch.

frontend/app/components/Home.tsx
```diff
@@ -1,9 +1,12 @@
...
-import { Alert, Grid, Typography } from '@mui/material';
+import React, { useState } from 'react';
+import { useQuery } from 'react-query';
+import { Alert, Box, Grid, LinearProgress, Typography } from '@mui/material';
 import { styled } from '@mui/material/styles';
 
+import ErrorsDisplay from './ErrorsDisplay';
 import HeaderAppBar from './HeaderAppBar';
 import ToolCard from './ToolCard';
+import * as api from '../api';
```
jonespm, line 9: Maybe nitpick but Airbnb style guide says [do not use wildcard imports](https://github.com/airbnb/javascript#modules--no-wildcard). This this really needed here? Is this common for some situations?

frontend/app/components/Home.tsx
jonespm: Looks like a good, well supported library

jonespm (approved):

frontend/app/components/Home.tsx
ssciolla: This is a pattern I followed from ROHQ. It's kind of nice to see in this case which functions are coming from the `api` module (and which by extension depend on external connections), but I don't mind changing it.

frontend/app/api.ts
ssciolla: Good question. I have fallen into the habit of using `fetch` in the browser, probably because it's built in and our other projects have done this. I have used `axios` and `got` in CCM though, so some familiarity with the alternatives. I did find this article comparing `axios` to `fetch` for client-side work: https://blog.logrocket.com/axios-vs-fetch-best-http-requests/ `axios` is pretty small, so I'm not sure "keeping the bundle size small" is a super convincing argument. I think if we add more endpoints in the future, it would be worth considering switching to something else. Biggest benefit to switching right now I'm seeing would be not having to `stringify` the body, or `response.json` to get the response data.

ssciolla: I created an issue to track this suggestion, perhaps we'll want to explore it in the future. #206

frontend/app/components/Home.tsx
ssciolla: OK, changed it, see https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/196/commits/124b5de9e40f5a9931e7c9745f7a515b6e18ab4b

frontend/app/components/ErrorsDisplay.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Alert, Grid } from '@mui/material';
+
+interface ErrorsDisplayProps {
+  errors: Error[]
+}
+
+export default function ErrorsDisplay (props: ErrorsDisplayProps) {
+  return (
+    props.errors.length === 0
+      ? <></>
```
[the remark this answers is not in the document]
ssciolla: I just modified this component to not check the length. As it's implemented, it's still type safe; the worst that could happen is someone uses it with an empty array, and there might be a little leftover spacing (would have to test that to be sure). I also think we can expect that the user (us) to check the length and only use it if the array has one or more errors.

ssciolla:
FYI, I fixed a styling-related issue, resulting in some extra empty space on the tool cards. Should be resolved now. See https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/196/commits/b82eab28028207d64047e6929bd30fcaa0e5c24c

zqian:
Not sure where the 'extra empty space' was. [b82eab2]( https://github.com/tl-its-umich-edu/canvas-app-explorer/commit/b82eab28028207d64047e6929bd30fcaa0e5c24c) looks good to me.

PR #199, opened 2022-06-13
Add a dialog for viewing larger versions of screenshots (#185)

### LINKED ISSUE #185: Allow for larger version of tool screenshot in dialog
UX team feedback adapted from meeting notes

"Main images in the expand-able area are very small. Should we add a dialog that pops up to make it easier to see the screenshots? ... I do agree there needs to be a way to enlarge the image. I’m not sure what is easier to implement. 1) Use of a dialog. 2) A way to zoom in"

ssciolla:
This PR aims to resolve #185.


frontend/app/components/ToolCard.tsx
```diff
@@ -30,32 +27,21 @@ export default function ToolCard (props: ToolCardProps) {
...
-  }
-
-  let mainImageProps: CardImageProps = {
-    component: 'img',
-    height: 150,
-    alt: tool.main_image_alt_text ?? 'Image of tool in use'
-  };
-  if (tool.main_image !== null) {
-    mainImageProps = { image: tool.main_image, ...mainImageProps };
-  }
```
ssciolla, line 50: I commented in an earlier PR that I didn't like how I did this. I found while working on this that `CardMedia` component was easier to use (and probably more readable) when I just used things inline. I forget why I did not do that earlier.

frontend/app/components/toolButtons.tsx
```diff
@@ -1,12 +1,7 @@
 import React from 'react';
 import AddIcon from '@mui/icons-material/Add';
 import RemoveIcon from '@mui/icons-material/Remove';
-import { Button, ButtonProps, SvgIconProps, Typography } from '@mui/material';
-
-const iconProps: SvgIconProps = {
-  fontSize: 'medium',
-  sx: { marginRight: 1 }
-};
```
ssciolla, line 9: I realized that `medium` is the default, and just decided to inline this, since that's how I'm doing it elsewhere. Maybe the icon and text combination with buttons can be extracted some way in the future, but this seems fine for now.

ssciolla: I also made this change and the others below in #200, I'll have to rebase depending on what gets merged first.

ssciolla: Okay, I rebased and the PR should be limited to the dialog changes now.

frontend/app/components/ToolCard.tsx
```diff
@@ -80,7 +66,43 @@ export default function ToolCard (props: ToolCardProps) {
...
+            <DialogActions>
+              <Button
+                aria-label='Close screenshot dialog'
+                onClick={() => setScreenshotDialogOpen(false)}
+                startIcon={<CloseIcon />}
+              >
+                Close
+              </Button>
+            </DialogActions>
+          </Dialog>
```
ssciolla, line 105: A thought: I think it's reasonable to expect that every tool that we post information about has a `logo_image`, but it's possible we may want to go live with tools without a screenshot. Should we make all of this conditional, hiding it when `tool.main_logo` is not defined? Thoughts? This could be done now or in a separate issue.

zqian: That's good idea. As of now, neither the "Logo image" or the "Main image" is of required field, so either could be null.

Is it possible to also make the image itself clickable, so that once the user clicks on the main image (if exists), the dialog window will be shown, the same user experience as if the clicked on the "Enlarge Screenshot" button.

frontend/app/components/ToolCard.tsx
```diff
@@ -30,24 +27,7 @@ export default function ToolCard (props: ToolCardProps) {
...
-    alt: tool.logo_image_alt_text ?? 'Logo for tool'
-  };
-  if (tool.logo_image !== null) {
-    logoImageProps = { image: tool.logo_image, ...logoImageProps };
-  }
-
-  let mainImageProps: CardImageProps = {
-    component: 'img',
-    height: 150,
-    alt: tool.main_image_alt_text ?? 'Image of tool in use'
```
zqian, line 46: Can the alt text contain tool name, such as "Screenshot of Gradecraft tool"

frontend/app/components/ToolCard.tsx
```diff
@@ -30,24 +27,7 @@ export default function ToolCard (props: ToolCardProps) {
   const { tool } = props;
 
   const [showMoreInfo, setShowMoreInfo] = useState(false);
-
-  let logoImageProps: CardImageProps = {
-    component: 'img',
-    height: 150,
-    alt: tool.logo_image_alt_text ?? 'Logo for tool'
```
zqian, line 37: Can the alt text include tool name, such as "Logo image for Gradecraft tool"?

frontend/app/components/ToolCard.tsx
ssciolla: Sure, I'll add that

frontend/app/components/ToolCard.tsx
ssciolla: Sure

ssciolla: Should be changed now.

frontend/app/components/ToolCard.tsx
ssciolla: Should be changed now.

frontend/app/components/ToolCard.tsx
zqian: verified!

jonespm:
It looks like Zhen had good feedback and this looks good to me. Do you know how this looks if the image is **too** big, like if someone uploads a 4000px image? I think we'll have to implement the fix for resizing to get both images and thumbnails sooner.
https://picsum.photos/4000

ssciolla:
> Do you know how this looks if the image is **too** big, like if someone uploads a 4000px image? I think we'll have to implement the fix for resizing to get both images and thumbnails sooner.

@jonespm, do we have an issue open for this? I did test this just now, it loads pretty slow, with some amount of the image loading in strips. Re-sizing would make sense, or some other image handling mechanisms the whole image appears at once, even if delayed.

PR #204, opened 2022-06-16
Re-design `Header.tsx` as `HeaderAppBar.tsx` (#182)

### LINKED ISSUE #182: Consider using `@mui.AppBar` with or in place of `Header.tsx`
Do you think we might be use for a more standard AppBar style component here for the header? 
https://mui.com/material-ui/react-app-bar/

We do need a admin button for admins. And it would probably be useful to have a profile icon. Something more like this?
https://mui.com/material-ui/react-app-bar/#app-bar-with-search-field

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/171#discussion_r887189974_

ssciolla:
This PR aims to resolve #182. Header sizes were also modified, so the `h1` uses `h5` styling and the `h2` uses `h6`.


frontend/app/components/HeaderAppBar.tsx
```diff
@@ -0,0 +1,42 @@
...
+export default function HeaderAppBar (props: HeaderAppBarProps) {
+  return (
+    <AppBar position='sticky'>
+      <Toolbar>
+        <Grid container direction='row' alignItems='center'>
+          <Grid item xs={6}>
+            <Typography variant='h5' component='h1' sx={{ marginRight: 1 }}>Canvas App Explorer</Typography>
+          </Grid>
+          <Grid
+            item
```
jonespm, line 18: Maybe nitpick, but while looking at this I noticed sometimes there's many attributes on a single like (like Typography on line 15 and Grid on line 13) and other times only one attribute per line spanning over 10 lines. Is this some type of recommended coding style for React or TS?

I understand keeping line lengths short but this seems like it's resulting in a lot of horizontal scrolling for not much purpose. I _feel_ like 3 attributes per line would generally be a good compromise. Is an auto-formatter doing this?

jonespm (approved):
Aside from the comment about attributes per line, this looks like a good change!

frontend/app/components/HeaderAppBar.tsx
ssciolla: Nope, no auto formatter at play here. What I'm doing seems to me pretty common in the React community -- put all props on one line unless it gets too long, then one per line when it goes to multi-line. See the Airbnb guide which is popular: https://github.com/airbnb/javascript/tree/master/react#alignment I may need to split some of these into multiple lines.

Looks like there are eslint rules you can use, I'll have to check whether they are in place.

jonespm: Thanks! I hadn't seen this before. It seems like it makes it harder to read with a lot of small props on a single line to me, but I'm fine if there's a style-guide you're going off of. 

Maybe it's just because they seem randomly ordered, and [perhaps sorting](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md) would help but I agree with that page that it _could_ make it harder to maintain. It doesn't sound like they have any suggested practice for that.

ssciolla: FWIW, I don't follow the Airbnb guide closely; most of my style practices are a combination of things I've picked from working with `standard` or learned from others like Justin or James. At some point we could come up with a style guide and enforce it with `eslint` rules, but that seems low priority at this point.

jonespm: Yeah, I think that would help save time if anyone else ever contributed to the frontend and to keep the style consistent. 

I see that AirBnb has a [package for their rules](https://www.npmjs.com/package/eslint-config-airbnb). Developers might either be too intimidated to contribute to the frontend or frustrated with all the nitpicking in PR's if most comments were about style. I personally would feel a little of both and prefer the majority of easy issues be picked up by the IDE.

Issue #208, opened 2022-06-20
If refresh token is invalid, rather than returning invalid_grant, just remove token and retry


jonespm:
I believe that when calling this [get_oauth_token](https://github.com/tl-its-umich-edu/canvas-app-explorer/blob/7c877a5c8fb5388481556ef86b3330ec2c068dd2/backend/views.py#L7) code if the access token is invalid it ends up printing
`Error: refresh_token request failed to get a token: Error: invalid_grant`

We'd ideally want it to try to remove this users token and try the process again. It's possible that something changed that caused the token to become invalid and clearing it automatically would fix it. 

I'm not 100% sure where the error is pushed up from but we'd have to investigate a little and hopefully can take care if it in our code rather than the library.

To reproduce
* Launch the app and get a token.
* Go into the Admin interface and click on Canvas Oauth2 Tokens
* Click on the token for your user id
* Change the refresh token and update the date to be "now" so it invalidates it
* Try to launch again. You should get this error.

Expectation is that it does one refresh to try to clear out and retry rather than just fail.

We may also want to test the case if the user removes their token from Canvas, it should be able to handle this. 

This workaround is currently referenced at the [bottom of the wiki](https://github.com/tl-its-umich-edu/canvas-app-explorer/wiki/Configuring-LTI-1.3) with a manual workaround.

zqian:
I think this error after the Canvas test instance reset, it invalidate the old Canvas OAuth2 token?

jonespm:
Yes, this will also happen after every Canvas test refresh. I'm not sure if there's other cases it could happen. We haven't tested many exception cases for OAuth.

roujiawang:
Performed tests according to the TEST PLAN
No unexpected behavior found.

PR #211, opened 2022-06-22
issue_208 remove invalide refresh token and redirect user to the OAuth page for new token

zqian:


ssciolla:
I'll take a look at this after lunch.

ssciolla:
So this is focused on the refresh token, but there are other errors that might occur (when an authorization is removed on the Canvas side) before the token needs to be refreshed. Are we going to handle that separately?

zqian:
Right. This PR only addresses the invalid refresh token problem as reported in #208 

If the authorization is removed from Canvas User profile settings, the error message will be "Invalid access token", and that should be reported as a different issue.

backend/views.py
```diff
@@ -1,8 +1,22 @@
-from canvas_oauth.oauth import get_oauth_token
+from canvas_oauth.oauth import get_oauth_token, handle_missing_token
+from canvas_oauth.models import CanvasOAuth2Token
 from django.contrib.auth.decorators import login_required
 from django.shortcuts import render
+from django.contrib.auth.models import User
+import logging
+
+logger = logging.getLogger(__name__)
```
ssciolla, line 8:
```suggestion
import logging

from canvas_oauth.oauth import get_oauth_token, handle_missing_token
from canvas_oauth.models import CanvasOAuth2Token
from django.contrib.auth.models import User
from django.contrib.auth.decorators import login_required
from django.shortcuts import render



logger = logging.getLogger(__name__)
```

backend/views.py
```diff
@@ -1,8 +1,22 @@
...
-    access_token = get_oauth_token(request)
+    try:
+        access_token = get_oauth_token(request)
+    except Exception as err:
+        logger.error(err)
+        if 'refresh_token request failed to get a token' in str(err):
+            logger.error("removing invalid refresh_token")
+            CanvasOAuth2Token.objects.filter(user=request.user).delete()
+        # prompt for new OAuth token
+        return handle_missing_token(request)
```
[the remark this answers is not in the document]
zqian: Good idea. I have added more detailed OAuth exception handlings, so that

- MissingTokenError is caught after user's Canvas OAuth2 Token is removed from Django admin page
- InvalidOAuthStateError is caught after user's Canvas OAuth2 refresh Token is modified to invalid value

Both exceptions are handled by reauthentication.

backend/views.py
```diff
@@ -1,8 +1,36 @@
...
+        logger.error("InvalidOAuthStateError. Remove invalid refresh_token and prompt for reauthentication.")
+        CanvasOAuth2Token.objects.filter(user=request.user).delete()
+        return handle_missing_token(request)
+    except InvalidOAuthReturnError as err:
+        logger.error("InvalidOAuthReturnError. Remove invalid refresh_token and prompt for reauthentication.")
+        CanvasOAuth2Token.objects.filter(user=request.user).delete()
+        return handle_missing_token(request)
+    except MissingTokenError as err:
+        logger.error("MissingTokenError. Prompt for reauthentication.")
+        return handle_missing_token(request)
```
ssciolla, line 28: I **think** this `except` is unnecessary, since if the `MissingTokenError` occurs, the Canvas OAuth middleware will handle the redirect to reauthenticate -- see https://github.com/Harvard-University-iCommons/django-canvas-oauth/blob/6a2dcff2474f4b7bc7367f41d70e0bbb377ee6c0/canvas_oauth/oauth.py#L18-L45

backend/views.py
```diff
@@ -1,8 +1,36 @@
...
+
 @login_required
 def get_home_template(request):
-    access_token = get_oauth_token(request)
+    try:
+        access_token = get_oauth_token(request)
+    except InvalidOAuthStateError as err:
+        logger.error("InvalidOAuthStateError. Remove invalid refresh_token and prompt for reauthentication.")
+        CanvasOAuth2Token.objects.filter(user=request.user).delete()
+        return handle_missing_token(request)
```
ssciolla, line 21: This one may also be unnecessary, since this is part of the original OAuth handshake and relates to the state parameter and prevention of CSRF. I don't think we'd want to delete the token in this case.

backend/views.py
```diff
@@ -1,8 +1,36 @@
...
+    except MissingTokenError as err:
+        logger.error("MissingTokenError. Prompt for reauthentication.")
+        return handle_missing_token(request)
+    except CanvasOAuthError as err:
+        logger.error("CanvasOAuthError. Remove invalid refresh_token and prompt for reauthentication.")
+        CanvasOAuth2Token.objects.filter(user=request.user).delete()
+        return handle_missing_token(request)
+    except Exception as err:
+        logger.error(err)
```
ssciolla, line 34: I think we could just let other errors not be caught, and let the library's middleware handle them. What do you think? https://github.com/Harvard-University-iCommons/django-canvas-oauth/blob/master/canvas_oauth/middleware.py

ssciolla: In summary, we could just have a `try` / `except` catching `InvalidOAuthReturnError`, and leave the rest out.

zqian: That sounds good. I will leave the InvalidOAuthReturnError handling and remove others.

backend/views.py
zqian: removed this error handling.

backend/views.py
zqian: Removed the MissingTokenError error handling section.

backend/views.py
```diff
@@ -1,8 +1,23 @@
...
+
+logger = logging.getLogger(__name__)
+
 @login_required
 def get_home_template(request):
-    access_token = get_oauth_token(request)
+    try:
+        access_token = get_oauth_token(request)
+    except InvalidOAuthReturnError as err:
+        logger.error("InvalidOAuthReturnError. Remove invalid refresh_token and prompt for reauthentication.")
```
jonespm, line 19: This error message really doesn't contain any details that would be useful and would just be noise in the logs. Maybe it could at least include `request.user` incase we needed to try to debug an issue.

zqian: Done. Added {request.user}. check [cca944e](https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/211/commits/cca944e22591fdd7b13e2a1b2027f7714138fe22)

backend/views.py
```diff
@@ -1,8 +1,23 @@
...
 @login_required
 def get_home_template(request):
-    access_token = get_oauth_token(request)
+    try:
+        access_token = get_oauth_token(request)
+    except InvalidOAuthReturnError as err:
+        logger.error(f"InvalidOAuthReturnError for user: {request.user}. Remove invalid refresh_token and prompt for reauthentication.")
+        CanvasOAuth2Token.objects.filter(user=request.user).delete()
+        return handle_missing_token(request)
+
```
ssciolla, line 22: Nit-pick: there are a couple unused imports/variables here: `User`, `access_token`, and `err`.

zqian: Done. [9c7c2b9](https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/211/commits/9c7c2b9af2d7b5b259bffe79ea88d7e433f4c8ea)

PR #214, opened 2022-06-22
Add Canvas error details and standardize API error body (#205)

### LINKED ISSUE #205: Pass Canvas error information to the Django API error response
I think this PUT call invokes the LTIToolViewSet>update function inside Views.py. In line 57 of Views.py, where the Response is returned in Exception handling block:

```
try:
	manager.update_tool_navigation(canvas_id_num, not navigation_enabled)
except Exception as err:
	logger.error(err)
	return Response(status=500)
```

Could the Response object also include the err message, or some other user friendly error message?

_Originally posted by @zqian in https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/196#discussion_r898655295_

ssciolla:
The PR aims to resolve #205. This is still a draft. Feedback is welcome.

With these changes, I believe many errors originating from our application and all from Canvas should have an error body like the following: 
```
{
  status_code: 404
  message: "An error occurred while communicating with Canvas: "Not Found"
}
```

TEST PLAN:

- Setup proper values for TEST_API_KEY, TEST_API_URL and TEST_COURSE_ID fields in .env
- Start CAE tool
- In command line, run the test cases `docker exec -it canvas_app_explorer /bin/bash -c "python manage.py test"`
- Observe the success test results. The `test_unauthorized_error` and `test_not_found_error` test cases actually tested for the error.status_code and error.message changes.

### DIFF
--- a/backend/canvas_app_explorer/canvas_lti_manager/manager.py
+++ b/backend/canvas_app_explorer/canvas_lti_manager/manager.py
@@ -48,7 +48,7 @@ class CanvasLtiManager:
         for class_key in EXCEPTION_STATUS_MAP:
             if isinstance(exception, class_key):
                 return CanvasHTTPError(exception.message, EXCEPTION_STATUS_MAP[class_key])
-        return CanvasHTTPError(exception.message, 500)
+        return CanvasHTTPError(exception.message, HTTPStatus.INTERNAL_SERVER_ERROR.value)
 
     def create_external_tool_tab(self, tab: Tab) -> ExternalToolTab:
         return ExternalToolTab(
--- a/backend/canvas_app_explorer/views.py
+++ b/backend/canvas_app_explorer/views.py
@@ -2,7 +2,7 @@ import logging
 
 from django.conf import settings
 from drf_spectacular.utils import extend_schema, OpenApiParameter
-from rest_framework import authentication, permissions, viewsets
+from rest_framework import authentication, permissions, status, viewsets
 from rest_framework.request import Request
 from rest_framework.response import Response
 
@@ -13,6 +13,7 @@ from backend.canvas_app_explorer.canvas_lti_manager.exception import CanvasHTTPE
 logger = logging.getLogger(__name__)
 
 MANAGER_FACTORY = DjangoCourseLtiManagerFactory(f'https://{settings.CANVAS_OAUTH_CANVAS_DOMAIN}')
+
 
 class LTIToolViewSet(viewsets.ViewSet):
     """
@@ -51,8 +52,10 @@ class LTIToolViewSet(viewsets.ViewSet):
         try:
             canvas_id_num = int(canvas_id)
         except ValueError:
-            bad_request_data = { 'status_code': 400, 'message': 'canvas_id must be an integer.' }
-            return Response(data=bad_request_data, status=400)
+            bad_request_data = {
+                'status_code': status.HTTP_400_BAD_REQUEST, 'message': 'canvas_id must be an integer.'
+            }
+            return Response(data=bad_request_data, status=status.HTTP_400_BAD_REQUEST)
         serializer = serializers.UpdateLtiToolNavigationSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
         navigation_enabled: bool = serializer.validated_data['navigation_enabled']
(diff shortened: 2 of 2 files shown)


backend/canvas_app_explorer/canvas_lti_manager/manager.py
```diff
@@ -1,17 +1,38 @@
...
+    ResourceDoesNotExist, Unauthorized, UnprocessableEntity
+)
 from canvasapi.tab import Tab
 
 from .data_class import ExternalToolTab
+from .exception import CanvasHTTPError
+
+
+EXCEPTION_STATUS_MAP = {
+    BadRequest: 400,
```
zqian, line 16: can we use value as coded from the [HTTPSTATUS library](https://docs.w3cub.com/python~3.9/library/http#http.HTTPStatus)? e.g. `BadRequest: HTTPStatus.BAD_REQUEST.value`

zqian:
Since this is a "wrapper" for canvasapi library, maybe we can provide this code and file an issue to UCF's canvasapi library?

backend/canvas_app_explorer/canvas_lti_manager/manager.py
ssciolla: This you mean? https://docs.python.org/3.8/library/http.html#http-status-codes That's a good idea, thanks

zqian: LGTM! Thx!

ssciolla: See https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/214/commits/958155f7b8effcdcd383123044f30f0d79974092.

ssciolla:
> Since this is a "wrapper" for canvasapi library, maybe we can provide this code and file an issue to UCF's canvasapi library?

Which part particularly? I think it would be nice if the library's exceptions included the status codes, is that what you mean?

zqian:
` I think it would be nice if the library's exceptions included the status codes, is that what you mean?`
Yes, that's what I meant.

backend/canvas_app_explorer/canvas_lti_manager/manager.py
zqian: Looks good!

zqian (approved):

jonespm:
Yeah, I was kind of thinking the same thing as @zqian when I was looking at this, but that may not be the quickest solution. Still probably worth filing an issue though they probably wouldn't fix it unless we did a PR.

backend/canvas_app_explorer/canvas_lti_manager/manager.py
```diff
@@ -22,6 +43,13 @@ def __init__(self, api_url: str, api_key: str, course_id: int):
         self.course_id: int = course_id
         self.canvas_api: Canvas = Canvas(api_url, api_key)
 
+    @staticmethod
```
jonespm, line 46: It seems like this method and all of this exception code before this could just be consolidated into to `exception.py`. Then this could be a method on the CavasHTTPError class and you could call something like error.convert_error().

Was there are reason for having it external to that in the manager.py here?

ssciolla: Even if we got the `canvasapi` library updated, I'm not sure I'd want to reference `CanvasException` or its attributes in `views.py`. I think that would defeat the benefits of having an adapter like `CanvasLTIManager` in the first place, which again are a) to make testing easier and b) make it easier to maintain because the manager class could get rewritten with another library. Any future classes could also crosswalk their exceptions to `CanvasHTTPException` (which I think is really just dealing with raw Canvas API data) with a conversion method too.

It would still be nice to have `canvasapi` provide the status codes, it'd mean we could remove the map dictionary I set up. I agree with you that it's not likely we'd need to replace this library in the short term, but I'm trying to learn and try out good/interesting code design practices.

ssciolla: I did want to hear your opinion on this, talk through my thinking, that's why I asked for your review 😁

jonespm: I think it's fine, not going to nitpick it. 

I think if I'd done it I'd just have put this `convert_error` as a method in `CanvasHTTPError` rather than in `CanvasLtiManager` here and call `error.convert()` (or even have it part of `__str__`) rather than` self.convert_error(error)`.

jonespm (approved):

Issue #212, opened 2022-06-22
Switch database for localhost from mysql-server back to original mysql


jonespm:
There was some things I noticed with the MySQL server switch we did on #186 

This is it switch back over to mysql and use `mysql:8-oracle`

-  It has a "warning" that it's built for Linux platforms and others aren't supported. They specifically mention Windows (and possibly OSX but it doesn't say that)
- I can't seem to get the root password to work and login with it still seeing if thats the same. I think it's because `MYSQL_ROOT_HOST` in mysql-server is more restrictive?
- I've been using `mysql:8-oracle` today and everything works plus the performance **seems** better

PR #215, opened 2022-06-22
Fixes #147 - Disable existing admin backend and trust LTI for users that are admin

### LINKED ISSUE #147: Disable old admin backend and trust LTI for users that are admin
We currently have a backend login for admin that we're using in the LTI version. When in LTI this should be disabled and there should be some link displayed for users who are super users in the database to be able to click on and access the admin interface. 

This link may need to be added to plasmic or maybe we can just put it in the corner. We currently don't have any menus in this app but maybe it could use a ☰ left of the "Canvas App Explorer" title and admin could be in there. Not sure what else though?

jonespm:
Because the backdoor to add an initial pylti and super user are removed from this. I'm planning on also updating the Wiki documentation. This significantly simplifies local setup.

### Run this command to create an initial PyLTI Key. These are the minimal arguments

These are the minimal values to set. You can run it multiple times to update a previous client_id+issuer pair. To remove you'll have to go into the admin UI or database.

`docker exec -it canvas_app_explorer ./manage.py manage_pylti --tool_key="LTI Key" --title="Canvas Test" --client_id=85530000000000126`

You can additionally specify 
`--platform` (Defaults to canvas.instructure.com)
`--deployment_ids` (Needs to be set unless the .env value LTI_CONFIG_DISABLE_DEPLOYMENT_ID_VALIDATION=True is set)
  You can specify multiple like 
  `--deployment_ids "303:43090cc0c11063fdf5b270094da05ed6e23ea6ad" "100:43090cc0c11063fdf5b270094da05ed6e23ea6ad"`
  These are obtained from the External Apps Page on Canvas

### After you set this up, login from Canvas and Authorize the app. You should see the home page.

Then run this to promote yourself or another logged in user initially to staff. (Fill in the UPDATE_USERNAME= part)

`UPDATE_USERNAME="jonespm@umich.edu";docker exec -it canvas_app_explorer ./manage.py shell -c "from django.contrib.auth.models import User; User.objects.update_or_create(username='$UPDATE_USERNAME', defaults=dict(is_superuser=True, is_staff=True))"`

After the first time doing this you can login as a user and promote through the admin UI and the link.


jonespm:
#212 should be merged or rejected before this is merged as this includes that commit.

backend/urls.py
```diff
@@ -33,6 +33,10 @@
     path('api/', include(router.urls)),
     path('api/auth/', include('rest_framework.urls', namespace='rest_framework')),
 
+    # Redirect all of these to a login template page to display a message
+    path('accounts/login/', views.get_login_template, name='custom_login_view_accounts'),
+    path('admin/login/', views.get_login_template, name='custom_login_page_admin'),
+    path('admin/logout/', views.get_login_template, name='custom_logout_view_admin'),
```
ssciolla, line 39: Is there a way to remove these rather than overriding them? Just curious. I saw you could customize the `AdminSite` class, but didn't see how Django itself recommends doing this. https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#adminsite-objects

backend/canvas_app_explorer/management/commands/manage_pylti.py
```diff
@@ -0,0 +1,63 @@
...
+
+# https://github.com/dmitry-viskov/pylti1.3/blob/master/pylti1p3/contrib/django/lti1p3_tool_config/models.py
+from pylti1p3.contrib.django.lti1p3_tool_config.models import LtiToolKey, LtiTool
+
+
+class Command(BaseCommand):
+    help = "Generate a private/public key pair and add them to the pylti13 database. This can be called with the same issuer + client ID to update a previous key. The Tool Key will be re-used"
+
+    def add_arguments(self, parser: CommandParser):
+        parser.add_argument('--auth_login_url', dest='auth_login_url', type=str, help="Canvas Login URL",
```
zqian, line 17: should those canvas urls be configurable for test vs production Canvas server urls?

backend/views.py
```diff
@@ -1,8 +1,14 @@
...
 from django.shortcuts import render
 
 @login_required
 def get_home_template(request):
     access_token = get_oauth_token(request)
     return render(request, 'home.html')
+
+
+def get_login_template(request):
+    messages.error(request, "You need to launch this tool from a LTI tool provider.")
```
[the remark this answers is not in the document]
zqian: maybe instead of name get_login_template, can we use get_error_template instead, and pass in error message as parameter?

backend/urls.py
jonespm: I didn't see any ideas either. this [admin.site.urls](https://github.com/django/django/blob/c627226d05dd52aef59447dcfb29cec2c2b11b8a/django/contrib/admin/sites.py#L317) is a dynamic method that generates urls from get_urls. 

It would be possible to override/extend some of these classesthat but I don't know if that's a better idea than this.

The URLS [are here.](https://github.com/django/django/blob/c627226d05dd52aef59447dcfb29cec2c2b11b8a/django/contrib/admin/sites.py#L269)

There are some others here like /admin/password-change but they redirect to /admin/login first.

I basically [used this suggestion from here](https://stackoverflow.com/a/61946707/3708872) without finding a better one. 

I think the eventual fix would be that we write our custom admin, I just wanted to remove any potential backdoors.

jonespm: There were some other ideas here. Rather than define a template with a nice message we **could** use the view [django.views.defaults.page_not_found](https://docs.djangoproject.com/en/4.0/ref/views/#the-404-page-not-found-view).

I also didn't try if it's possible to do something like add the path
`    path('admin/', admin.site.urls),`

But then remove these paths from the list?

README.md
```diff
@@ -11,14 +11,7 @@ Then the app in development should be accessible on http://localhost:5000/
...
 
-To create a local admin user you should run this.
-
-```
-docker exec -it canvas_app_explorer ./manage.py shell -c "from django.contrib.auth.models import User; User.objects.create_superuser('admin', 'admin@example.com', 'admin')"
-```
-
-Please see the Wiki for instructions on configuring with LTI.
-
+This app can currently only be launched via LTI.  Please see the Wiki for instructions on configuring with LTI.
```
zqian, line 14: Since the admin role is solely defined in CAE Django database, we may need to add an instruction line here. 

Can include a sql update statement too: `update auth_user set is_staff=1 where id=x;`

backend/canvas_app_explorer/management/commands/manage_pylti.py
ssciolla: Not sure if this is exactly what @zqian is saying, but I was thinking you could just have the user provide the platform URL, and have the endings be configurable via environment variables with good defaults. I think this is fine, but it would make it a lot easier for test configuration if you just had to provide `--platform=canvas.test.instructure.com` and instead of entering all these `_url` flags on the command line.

README.md
jonespm: I put this in the instructions in the notes how you can update this with the application. That's **possibly** better than requesting the user login and run SQL. Basically is the same action. I figured this would just be in the wiki. 

`UPDATE_USERNAME="jonespm@umich.edu";docker exec -it canvas_app_explorer ./manage.py shell -c "from django.contrib.auth.models import User; User.objects.update_or_create(username='$UPDATE_USERNAME', defaults=dict(is_superuser=True, is_staff=True))"`

backend/canvas_app_explorer/management/commands/manage_pylti.py
jonespm: All of these are optionally configurable.

It doesn't **seem** to matter on test/dev/prod which ones you use. You probably should use the one for your server's URL but it works for me even with these defaults, at least on dev. 

If it doesn't work you can add these command line arguments and run it again and it will update it. 

Yeah, maybe rather than making these all individually configurable (or optionally in addition to) we could just have like a platform URL as you suggest. 

Actually that's probably easier for this really and a good idea. I'll just remove all those and if `--platform` is defined we can use that, otherwise just use `canvas.instructure.com`.

ssciolla: I think dev uses `canvas.instructure.com` for the platform; not sure if that works for test or beta though.

zqian: This canvas.instructure.com setting also works for Canvas test server. I tested it with the integration of my localhost.

I am confused now. Is the platform value already remains unchanged as canvas.instructure.com?

jonespm: I haven't done extensive testing but I think the only thing that doesn't work and return the same results is the jwks URL like this https://canvas.instructure.com/api/lti/security/jwks

But we aren't using this anyway in this app so it doesn't matter. **Possibly** also the OAuth URL's but those are in the .env file. 

I think there's a chance that canvas.instructure.com could be down while our instance is still up and break our app. So for prod and test we probably want to use the domain matching the instance we're testing on.

jonespm: I'm simplifying all of this so you only have to pass platform. 

You can also go into the database and edit these too, but I've found it to be a lot easier to run a management command to get this provisioned that editing them in the db, especially for the first one.

jonespm: I updated the command to just take the platform arg

backend/views.py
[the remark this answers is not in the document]
jonespm: Sure, I can change the name and pass it from here if I can't get a generic template to work better.

backend/urls.py
ssciolla: I think what you're doing is probably OK. Changing the name of the view was probably a good call.

jonespm: I looked into this and couldn't figure out anything great.

I have an idea of running it through a configurable filter. This **feels** like is should work. This produces a filtered tuple that looks like we can pass it into path. This can for sure be improved. 

```
from django.urls import URLPattern

def exclude_pattern(var, patterns: list):
    # Just ignore these if they don't have a name because it's probably a URLResolver
    if type(var) is URLPattern and var.name in patterns:
        return False
    return True

(list(filter(partial(exclude_pattern, patterns=["login", "logout", "password_change", "password_change_done"]), admin.site.urls[0])), admin.site.urls[1], admin.site.urls[2])
```

This looks like it's always a pattern of  [(list, tuple)](https://github.com/django/django/blob/c627226d05dd52aef59447dcfb29cec2c2b11b8a/django/urls/conf.py#L71).


I didn't like the default look of the page_not_found and it's not **really** a 404.

zqian (approved):

backend/canvas_app_explorer/management/commands/manage_pylti.py
```diff
@@ -0,0 +1,66 @@
...
+        parser.add_argument('--title', dest='title', required=True, type=str, help="LTI Title")
+        parser.add_argument('--tool_key', dest='tool_key', required=True, type=str,
+                            help="Name of Tool Key to use, will create if new")
+        parser.add_argument('--deployment_ids', dest='deployment_ids',
+                            nargs='*', type=str, help="List of Deployment ID(s). Can be multiple.", default="")
+
+    def handle(self, *args, **options: dict):
+        key: RsaKey = RSA.generate(4096)
+
+        self.stdout.write('Generating public and private key for LTI Tool Key table')
```
ssciolla, line 27: Maybe "LTI Tool Key record"?

backend/canvas_app_explorer/management/commands/manage_pylti.py
```diff
@@ -0,0 +1,66 @@
...
+                                         defaults=dict(title=options["title"],
+                                                       is_active=True,
+                                                       client_id=options["client_id"],
+                                                       issuer=issuer,
+                                                       use_by_default=False,
+                                                       auth_login_url=auth_login_url,
+                                                       auth_token_url=auth_token_url,
+                                                       key_set_url=key_set_url,
+                                                       tool_key=lti_key,
+                                                       deployment_ids=options["deployment_ids"]
```
ssciolla, line 65: `options['deployment_ids']` needs to be wrapped in a `json.dumps(...)` here. It didn't work for me when I used a single deployment ID, with or without wrapping `"`s around it (they get stripped by the parser I think).

backend/urls.py
```diff
@@ -33,6 +33,13 @@
     path('api/', include(router.urls)),
     path('api/auth/', include('rest_framework.urls', namespace='rest_framework')),
 
+    # Redirect all of these to a login template page to display a message
```
ssciolla, line 36:
```suggestion
    # Redirect all of these to an error template page to display a message
```

backend/canvas_app_explorer/management/commands/manage_pylti.py
ssciolla: Some other messages about the other steps may be appropriate.

backend/canvas_app_explorer/management/commands/manage_pylti.py
```diff
@@ -0,0 +1,66 @@
...
+            lti_key = LtiToolKey.objects.create(name=options["tool_key"],
+                                                private_key=key.exportKey().decode('utf-8'),
+                                                public_key=key.publickey().exportKey().decode('utf-8')
+                                                )
+        except IntegrityError:
+            # Key must already have been created
+            self.stdout.write(
+                f"Using existing key named {options['tool_key']}")
+            # Lookup the tool key
+            lti_key = LtiToolKey.objects.get(name=options["tool_key"])
```
ssciolla, line 51: I'm not sure how computationally expensive `RSA.generate` or these other methods are (it does seem to hang for a second), but it might make sense to use another pattern so you don't need to call these methods if it already exists. Something like this:
```
try:
    lti_key = LtiToolKey.objects.get(name=options['tool_key])
except LtiToolKey.DoesNotExist:
    key = RSA.generate(4096)
    lti_key = LtiToolKey.objects.create(
      name=options['tool_key'],
      private_key=key.exportKey().decode('utf-8'),
      public_key=key.publickey().exportKey().decode('utf-8')
    )
```
There's also a question of whether to use `queryset.exists()` or `queryset.get` with a `try`/`except`. Either probably is fine here, IDK.

backend/urls.py
```diff
@@ -33,6 +33,13 @@
...
+    path('accounts/login/', views.get_error_template, name='custom_login_view_accounts',
+         kwargs={'error_msg': "Account login is disabled as you need to launch this tool from an LTI tool provider"}),
+    path('admin/login/', views.get_error_template, name='custom_login_page_admin',
+         kwargs={'error_msg': "Admin login is disabled as you need to launch this tool from an LTI tool provider"}),
+    path('admin/logout/', views.get_error_template, name='custom_logout_view_admin',
+         kwargs={'error_msg': "Admin logout is disabled as you need to launch this tool from an LTI tool provider"}),
```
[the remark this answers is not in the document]
ssciolla: Nit-pick: You could also put periods at the end of the sentences.

ssciolla (changes requested):
Sorry for all the comments. I think I'd have given approval if the `deployment_ids` flag was functional, but there may be some other suggestions here you'll find worth considering. Let me know if you need clarification or disagree with anything.

requirements.txt
```diff
@@ -24,4 +24,7 @@ https://github.com/Harvard-University-iCommons/django-canvas-oauth/archive/v1.1.
 # For serving static files
 whitenoise==5.3.0
 
-canvasapi==2.2.0
\ No newline at end of file
+canvasapi==2.2.0
+
+# For generating LTI keys
+pycryptodome==3.14.1
```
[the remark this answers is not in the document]
ssciolla: Okay, that's fine, we can stick with this one.

backend/canvas_app_explorer/management/commands/manage_pylti.py
ssciolla: FYI, this is the only comment that blocked me from approving. I'll approve if you resolve it.

backend/urls.py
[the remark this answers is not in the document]
jonespm: All of these views could _probably_ have the same message. I put one in as the default. 

Or do you mean just to pass a partial like "Account Login" and "Admin Login" rather than the entire message? The get_error_template *might* be able to get the route then it'd doesn't need to be passed in separately and display like

`accounts/login is disabled as you need to launch this tool from an LTI tool provider.` I tried using default view like 
`defaults.page_not_found` but that would require us to provide a better 404 template page, otherwise there's no information on it. That seemed a little out of scope.

backend/canvas_app_explorer/management/commands/manage_pylti.py
```diff
@@ -0,0 +1,66 @@
...
+                                                       is_active=True,
+                                                       client_id=options["client_id"],
+                                                       issuer=issuer,
+                                                       use_by_default=False,
+                                                       auth_login_url=auth_login_url,
+                                                       auth_token_url=auth_token_url,
+                                                       key_set_url=key_set_url,
+                                                       tool_key=lti_key,
+                                                       deployment_ids=options["deployment_ids"]
+                                                       ))
```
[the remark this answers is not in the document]
jonespm: I was thinking of having some of this semi self contained so I could contribute it back to the pylti project. But this is really Canvas specific and we'd need different patterns for other LMSes. This was really just intended as an easier way than running SQL queries or editing the DB with an IDE to get this initial row in here that didn't currently exist. I was trying to figure this out for MyLA too so we could move the configuration there to the DB, but MyLA not as likely to be multi-tenant as this app will hopefully someday be. 

The only semi-annoying part of doing it without the script is generating the RSA keys. MyLA does have a utility for that but I usually just use the IMS website. I was just hoping this would save some time for local dev.

backend/views.py
```diff
@@ -1,8 +1,14 @@
...
 
 @login_required
 def get_home_template(request):
     access_token = get_oauth_token(request)
     return render(request, 'home.html')
+
+
+def get_error_template(request, error_msg="Error"):
+    messages.error(request, error_msg)
+    return render(request, "error_messages.html")
```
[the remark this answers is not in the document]
jonespm: I think this is simple enough to just leave it as a function based view to be consistent with the other function.

We may want to have a separate issue where we convert these to class based views if these are getting complicated. 

Using the `messages` is probably unnecessary here too though. Really I'm starting to lean toward the fix I came up with earlier to remove these paths entirely from the start and remove this error code entirely.

backend/canvas_app_explorer/management/commands/manage_pylti.py
```diff
@@ -0,0 +1,66 @@
+from Crypto.PublicKey import RSA
+from Crypto.PublicKey.RSA import RsaKey
+from django.core.management.base import BaseCommand, CommandParser
+from django.db.utils import IntegrityError
+
```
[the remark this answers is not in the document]
jonespm: I could see renaming it to create_pylti_tool_config. It really is specific to this library. 

I originally envisioned it as a command that could be used to CRUD the records in these PyLTI tables but right now it just can Create and Update.

backend/canvas_app_explorer/management/commands/manage_pylti.py
jonespm: I can just drop the table part, will add a few other messages

backend/canvas_app_explorer/management/commands/manage_pylti.py
jonespm: I'm rewriting it with this pattern.

backend/canvas_app_explorer/management/commands/manage_pylti.py
jonespm: Okay, this looks like it was saving it with single quotes instead of double. I switched to just use json.dumps as you suggested.

backend/urls.py
jonespm: So this seemed like a good idea but removing these results in an error when other code in admin tries to reverse these

`Reverse for 'login' not found. 'login' is not a valid view function or pattern name.`

[This solution to just wrap the method login in a login_required](https://stackoverflow.com/a/13186337/3708872) seems to work fine.

backend/canvas_app_explorer/management/commands/manage_pylti.py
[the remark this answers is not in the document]
jonespm: I think this could be done later if we write a test for this, but I'd like to try to get it back upstream. Just going to resolve for now. I refactored this since this change too.

backend/urls.py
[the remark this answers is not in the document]
jonespm: Removed this in 7cc0bc93a4a708751d317e604719f4e92dc3785a

backend/views.py
```diff
@@ -1,4 +1,5 @@
 from canvas_oauth.oauth import get_oauth_token
+from django.contrib import messages
```
ssciolla, line 2:
```suggestion
```

backend/canvas_app_explorer/management/commands/manage_pylti.py
```diff
@@ -0,0 +1,66 @@
+import json
+from Crypto.PublicKey import RSA
+from Crypto.PublicKey.RSA import RsaKey
+from django.core.management.base import BaseCommand, CommandParser
+from django.db.utils import IntegrityError
```
ssciolla, line 5: I think `RsaKey` and `IntegrityError` are unused?

backend/canvas_app_explorer/management/commands/manage_pylti.py
```diff
@@ -0,0 +1,66 @@
...
+
+# https://github.com/dmitry-viskov/pylti1.3/blob/master/pylti1p3/contrib/django/lti1p3_tool_config/models.py
+from pylti1p3.contrib.django.lti1p3_tool_config.models import LtiToolKey, LtiTool
+
+
+class Command(BaseCommand):
+    help = "Generate a private/public key pair and add them to the pylti13 database. This can be called with the same issuer + client ID to update a previous key. The Tool Key will be re-used"
```
ssciolla, line 12: Please check this help string for accuracy, clarity. Sounds initially like it just generates the key pair, but it does more than that.

backend/urls.py
```diff
@@ -25,14 +26,17 @@
 from backend.canvas_app_explorer import views as canvas_app_explorer_views #type: ignore
 
 from . import views
+
 # This is for Django Rest Framework
 router = routers.DefaultRouter()
 router.register(r'lti_tools', canvas_app_explorer_views.LTIToolViewSet, basename='ltitool')
 
+# Require being logged in to get to the admin page
+admin.site.login = login_required(admin.site.login)
```
ssciolla, line 35: So a couple things I noted, some of which may be commented on in the SO thread:
1) This doesn't disable `logout`, but it doesn't break if you use it on the Admin site. I don't see this as a huge issue, but thought I'd raise it as a discussion point.
2) You can still somewhat indirectly surface the login page via redirect. You can observe this if you login as a non-admin user, view the frame source, then navigate to `/admin`. You are logged in, so it doesn't block you from visiting the page, but it knows you don't have the proper permissions and lets you log in as a different user.

ssciolla (approved):
I'm approving, as I don't think we need another full review cycle, but I'm providing some additional comments for consideration.

backend/canvas_app_explorer/management/commands/manage_pylti.py
jonespm: Thanks, I re-read this and updated it to describe a little more what it's doing.

backend/urls.py
jonespm: Yeah, I noticed that if you call that it redirects to login which is disabled. 

If someone did figure out the admin and tried to login, they wouldn't be able to use the page. I don't think this is a risk or anything that's going to be mentioned. It's mostly to just block any potential external login attempts. 

It looks like I probably could just use user_passes_test instead to block it entirely.

`admin.site.login = user_passes_test(False, admin.site.login)`

PR #225, opened 2022-06-24
Use `Snackbar`, update text for refresh alert (#224)

### LINKED ISSUE #224: Update text, increase visibility of refresh alert when tools are updated
Per feedback from the UX team, we need to update the text of the refresh alert to be "Refresh the page to update the tool's appearance in the left-hand navigation." Ideally, we also need to do something to make the alert more noticeable (probably through a snackbar/toast) if the user has scrolled down to where the top of the page isn't visible.

ssciolla:
This PR aims to resolve #224.

### DIFF
--- a/frontend/app/components/Home.tsx
+++ b/frontend/app/components/Home.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 import { useQuery } from 'react-query';
-import { Alert, Box, Grid, LinearProgress, Typography } from '@mui/material';
+import { Alert, Box, Grid, LinearProgress, Snackbar, Typography } from '@mui/material';
 import { styled } from '@mui/material/styles';
 
 import ErrorsDisplay from './ErrorsDisplay';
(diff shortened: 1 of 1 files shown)


frontend/app/components/Home.tsx
```diff
@@ -102,6 +99,11 @@ function Home (props: HomeProps) {
       <Typography component='footer' sx={{ textAlign: 'center' }}>
         Copyright © 2022 The Regents of the University of Michigan
       </Typography>
+      <Snackbar open={showRefreshAlert} onClose={handleRefreshAlertClose} autoHideDuration={10000}>
+        <Alert severity='success' elevation={2} onClose={handleRefreshAlertClose}>
+          Refresh the page to update the tool&apos;s appearance in the left-hand navigation.
```
jonespm, line 104: Is this the only alert we are needing able to handle? I'm mostly thinking if you'd use a Snackbar instead of the ErrorsDisplay components also? Looks good otherwise but I haven't tested it out.

ssciolla: The other alerts are located within the card, or at the top of the page if the data fails to load. So I don't think there's the same usability problem, where the alert won't necessarily be visible.

That said, I take your point that it may be preferable to use alerts in a consistent manner. Do we think that errors should appear at the bottom of the page, and then maybe disappear?

frontend/app/components/Home.tsx
```diff
@@ -102,6 +99,11 @@ function Home (props: HomeProps) {
       <Typography component='footer' sx={{ textAlign: 'center' }}>
         Copyright © 2022 The Regents of the University of Michigan
       </Typography>
+      <Snackbar open={showRefreshAlert} onClose={handleRefreshAlertClose} autoHideDuration={10000}>
```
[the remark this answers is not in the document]
zqian: Got it working!

zqian (approved):

frontend/app/components/Home.tsx
ssciolla: I'm gonna merge, but let me know what you think about this. If we did make the error messages use a snackbar, we could also use something like `notistack` as we did with CCM. We can open a separate issue if you think it's worth it.

PR #234, opened 2022-07-01
Add tool name to card under the logo image (#229)

### LINKED ISSUE #229: Add tool name to card
While adding more tools, we found it is difficult in practice to find unique logos with good text for each tool. Thus, we've determined that we should add the tool's name to the card to help with clarity. This should probably be an `h3`.

ssciolla:
This PR aims to resolve #229.

### DIFF
--- a/frontend/app/components/ToolCard.tsx
+++ b/frontend/app/components/ToolCard.tsx
@@ -81,7 +81,7 @@ export default function ToolCard (props: ToolCardProps) {
       variant='outlined'
       sx={{ padding: 1, width: 328, borderColor: 'primary.main', borderWidth: '3px' }}
     >
-      <CardContent sx={{ height: 225 }}>
+      <CardContent sx={{ height: 260 }}>
         <CardMedia
           component='img'
           height={150}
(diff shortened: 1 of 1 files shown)


jonespm:
I wonder if we could just reduce the height of the logo a little rather than increasing the card height? Maybe Janel has some feedback on that? 

Our logos images do seem a little larger (296x150) than ones on comparable sites like Wolverine Access (48x48) and EduAppCenter (200x104) and Google Play (56x56)

Going from 150 to 115 doesn't seem like that big of a loss but it does leave more whitespace on the sides of the images.

ssciolla:
> I wonder if we could just reduce the height of the logo a little rather than increasing the card height?

Do you want me to change that here? Somewhat out of scope. I agree with what you said, they are rather large, but there's the potential for whitespace on the sides if the images aren't sized or cropped just right.

jonespm:
We probably don't need to change it here, just noticed it and saw the cards getting bigger. Can create a new issue and ask about it at the next meeting. Can just reduce the image and card height in that fix if we feel they're too big.

PR #236, opened 2022-07-01
Fixes #220 Replace X_FRAME_OPTIONS with CSP

### LINKED ISSUE #220: Replace X_FRAME_OPTIONS with CSP
X_FRAME_OPTIONS was only intended to be used as a quick temporary test. We should [replace that with CSP similar to MyLA.](https://github.com/jonespm/student-dashboard-django/commit/69917e4580be040aaea3a7e3b372176dd86827d3)

https://github.com/tl-its-umich-edu/canvas-app-explorer/blob/210fe85bc49e0e8a8e7051430931b650fa1111d8/backend/settings.py#L151

jonespm:
I was looking to make this config a little simpler based on what you had in CCM and only configure what is really necessary which is the FRAME_ANCESTORS. I could make the others configurable with this as a default but it's a little messier using environment variables over JSON for configuration.

To test this you'd want to be browsing with the console window open to see if there's any CSP errors reported. All I saw are errors from Canvas about Sentry and some warnings about future versions of TinyMCE which are part of the plugin.

Perhaps we could get rid of some of the unsafe-inline, but would need to do a little more work, especially on scripts. And it might not be entirely possible? 

I also have some Google things in here that we're not using yet, but I'm sure we'll be adding Google Analytics to this to get some idea of usage.

### DIFF
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -234,7 +234,7 @@ if CSRF_COOKIE_SECURE:
     # Enables Proxies that set headers
     USE_X_FORWARDED_HOST = os.getenv('USE_X_FORWARDED_HOST', True)
 
-# Set CSP_FRAME_SRC to the yours Canvas domains
+# Set CSP_FRAME_SRC to the your Canvas domains
 CSP_FRAME_ANCESTORS = ["'self'",] + os.getenv('CSP_FRAME_ANCESTORS', '').split(',')
 # Allow inline scripts and Google assets. I don't think these need to be configurable
 CSP_SCRIPT_SRC = ["'self'", "'unsafe-inline'", "'unsafe-eval'", "www.google-analytics.com"]
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -236,11 +236,12 @@ if CSRF_COOKIE_SECURE:
 
 # Set CSP_FRAME_SRC to the your Canvas domains
 CSP_FRAME_ANCESTORS = ["'self'",] + os.getenv('CSP_FRAME_ANCESTORS', '').split(',')
-# Allow inline scripts and Google assets. I don't think these need to be configurable
-CSP_SCRIPT_SRC = ["'self'", "'unsafe-inline'", "'unsafe-eval'"]
+# This is currently unsafe-inline because of PyLTI scripts. This may be fixed in the future.
+CSP_SCRIPT_SRC = ["'self'", "https:", "'unsafe-inline'"]
 CSP_IMG_SRC = ["'self'", "data:"]
 CSP_FONT_SRC = ["'self'"]
-CSP_STYLE_SRC = ["'self'", "'unsafe-inline'"]
+# Allow inline styles. There are a few styles that come up in the report so it seems easier to just allow unsafe-inline here.
+CSP_STYLE_SRC = ["'self'", "https:", "'unsafe-inline'"]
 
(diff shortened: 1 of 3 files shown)


backend/settings.py
```diff
@@ -236,6 +234,14 @@
     # Enables Proxies that set headers
     USE_X_FORWARDED_HOST = os.getenv('USE_X_FORWARDED_HOST', True)
 
+# Set CSP_FRAME_SRC to the yours Canvas domains
+CSP_FRAME_ANCESTORS = ["'self'",] + os.getenv('CSP_FRAME_ANCESTORS', '').split(',')
+# Allow inline scripts and Google assets. I don't think these need to be configurable
+CSP_SCRIPT_SRC = ["'self'", "'unsafe-inline'", "'unsafe-eval'", "www.google-analytics.com"]
```
ssciolla, line 240: Is `unsafe-eval` necessary? Seems to work for me.

I don't think we have `unsafe-inline` set here for CCM, but it seems to be required. Do you know if this is because of how `pylti1p3` works? Or some other reason specific to this tool?

ssciolla (approved):
This is working for me. I had a couple questions and comment language suggestions. I'm not sure I would have future-proofed with the Google stuff (seems like it would be clearer if that stuff is added at the same time as the analytics implementation), but I'll defer to you on that.

backend/settings.py
jonespm: Yeah, it looks necessary for the pyltip3. It has inline scripts. I don't know if it's necessary for anything else. It might be possible to improve this with hashes or even better suggesting to the author to add a nonce to the script tags. Then the hashes wouldn't change if the code changes.

I've [never done this](https://content-security-policy.com/hash/) but can look into seeing how much work it would be. 

I'm not sure if `unsafe-eval` is needed.

jonespm:
I'm fine removing the Google stuff for now. I'll create an issue (if there isn't one yet) about supporting Google Analytics and add this as a note of consideration to that.

backend/settings.py
jonespm: I've removed `unsafe-eval.` 

These scripts are in PyLTI, and I can't add them as a SHA because the script actually changes each time. (I noticed this as I kept on adding new SHA's on each reload and eventually gave up) 

It looks like you can use this [CSP library](https://django-csp.readthedocs.io/en/latest/nonce.html) to generate nonce for scripts and include them, but we'd have to also include this as a dependency in the PyLTI library for this to work or create some of our own classes or middleware? It looks like there's been some [desire to get this into Django Core](https://code.djangoproject.com/ticket/15727) but that's almost 11 years old. 

I'm not sure how else to deal with this the "best" way. The easiest is just to leave `unsafe-inline` there.

ssciolla: Not sure of the implications for `unsafe-inline` for scripts, but I'll trust your CSP expertise here. Penetration testing folks would also tell us if it's an issue.

zqian (approved):

backend/settings.py
jonespm: Yeah, we learned a little more from CCM and this but each app may have some differences in their security policy. We could simplify MyLA's though as very few of these options are ever necessary to be configured to setup the app.

Ideally we wouldn't have to allow unsafe-inline but we'd have to do more work around the packages we use to be able to remove that option. And it might need to be something supported by the framework like Django. 

I think since CSP passes that nonce as a request parameter we could submit a PR to have the PyLTI1p3 optionally use it. I just haven't had the time to work out a fix for that to test and submit to the author. Just low priority.

PR #237, opened 2022-07-05
Add icon with tooltip indicator of tool's inclusion in the navigation (#230)

### LINKED ISSUE #230: Add indicator that tool is already available in the course
During usability testing, some users were confused initially by the "Remove tool" button, not realizing the tool was already available. We could add some kind of indicator -- a badge, icon, etc. -- that communicates the tool is already available, installed, etc.

ssciolla:
The PR aims to resolve #230.

### DIFF
--- a/frontend/app/components/Home.tsx
+++ b/frontend/app/components/Home.tsx
@@ -7,6 +7,7 @@ import ErrorsDisplay from './ErrorsDisplay';
 import HeaderAppBar from './HeaderAppBar';
 import ToolCard from './ToolCard';
 import { getTools } from '../api';
+import { TOOL_MENU_NAME } from '../constants';
 import '../css/Home.css';
 import { Globals, Tool } from '../interfaces';
 
@@ -101,7 +102,7 @@ function Home (props: HomeProps) {
       </Typography>
       <Snackbar open={showRefreshAlert} onClose={handleRefreshAlertClose} autoHideDuration={10000}>
         <Alert severity='info' elevation={2} onClose={handleRefreshAlertClose}>
-          Refresh the page to update the tool&apos;s appearance in the left-hand navigation.
+          Refresh the page to update the tool&apos;s appearance in the {TOOL_MENU_NAME}.
         </Alert>
       </Snackbar>
     </div>
--- a/frontend/app/components/ToolCard.tsx
+++ b/frontend/app/components/ToolCard.tsx
@@ -1,10 +1,12 @@
 import React, { useState } from 'react';
 import AddBox from '@mui/icons-material/AddBox';
 import { useMutation } from 'react-query';
+import CheckCircleOutline from '@mui/icons-material/CheckCircleOutline';
 import ExpandLessIcon from '@mui/icons-material/ExpandLess';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import {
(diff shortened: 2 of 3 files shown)


frontend/app/constants.ts
```diff
@@ -0,0 +1,3 @@
+const TOOL_MENU_NAME = 'Course Navigation Menu';
```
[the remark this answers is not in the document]
ssciolla: Yes, I can do something like this.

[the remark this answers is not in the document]
ssciolla: https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/237/commits/dc45c362c4f244c65896e0324dab6e3e11e3851f

frontend/app/components/Home.tsx
```diff
@@ -7,6 +7,7 @@ import ErrorsDisplay from './ErrorsDisplay';
 import HeaderAppBar from './HeaderAppBar';
 import ToolCard from './ToolCard';
 import { getTools } from '../api';
+import { TOOL_MENU_NAME } from '../constants';
```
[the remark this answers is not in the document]
ssciolla: That's interesting, I honestly haven't implemented internationalization before. It doesn't look like with these libraries you'd have to extract all your text into variables -- mostly just tag them so language alternatives can be swapped in. Feel free to create an issue.

Mostly with this, I was just trying to keep certain language consistent across components. There's probably more we could do here.

frontend/app/components/ToolCard.tsx
```diff
@@ -128,6 +134,13 @@ export default function ToolCard (props: ToolCardProps) {
                 />
               )
           }
+          {
+            tool.navigation_enabled && (
```
zqian, line 138: Is it possible to add the tooltip to the "Add Tool" and "Remove Tool" button, so that maybe when people mouse over to those action buttons, they can also see the message?

ssciolla: This is something we could consider; I'm not sure yet of the benefit. Are you thinking that the "REMOVE TOOL" button would have the same tool tip the icon, or that the buttons would have different tool tips? I guess I'd be concerned with the UI getting cluttered, or repeating things unnecessarily. We could ask the UX team.

zqian (approved):

frontend/app/components/ToolCard.tsx
zqian: I was thinking within the , there are both and elements, so that same info text can be shown when user hover over the "Remove Tool" button, too.

Same setting likewise for the "Add Tool" button.

But I am OK with hearing from the UX team first.

ssciolla: Still not exactly clear. When you hover over or have in focus the add or remove button, where would the tooltip be and what would it say? On the button or on the icon?

jonespm (approved):

frontend/app/components/ToolCard.tsx
zqian: Sam and I had a discussion afterwards. We agree to wait for feedback from UX team first. Resolve this for now.

frontend/app/constants.ts
```diff
@@ -0,0 +1,6 @@
+const constants = Object.freeze({
+  toolMenuName: 'Course Navigation Menu'
+});
+
```
ssciolla, line 4:
```suggestion
```

PR #242, opened 2022-07-08
Fixes #241 - Database images aren't cached, add cache control

### LINKED ISSUE #241: Database images aren't cached, add cache control
I noticed that images that are loaded from the database aren't cached. This is going to put unnecessary load and extra traffic on us with every request

image

I [found this thread](https://github.com/victor-o-silva/db_file_storage/issues/103) here about how we can pass headers along to this url.

To test if this working, after the first reload all the images in the network tab should be coming from disk cache.

image

If a new image is uploaded the name is changed, so I think this caching for this [default recommended year is fine. ](https://medium.com/pixelpoint/best-practices-for-cache-control-settings-for-your-website-ff262b38c5a2)

jonespm:


backend/urls.py
```diff
@@ -41,7 +41,7 @@
     path('', views.get_home_template, name = 'home'),
 
     # For file storage
-    re_path(r'^files/', include('db_file_storage.urls')),
+    re_path(r'^files/', include('db_file_storage.urls'), {'extra_headers': {'Cache-Control': 'public, max-age=31536000'}}),
```
zqian, line 44: Can you translate 31536000 into hours/days?

zqian (approved):
Verified the caching works. Left a question for cache age.

backend/urls.py
jonespm: It's basically a year, what's generally suggested as a "constant" on various places

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#caching_static_assets_with_cache_busting

https://ashton.codes/set-cache-control-max-age-1-year/

If new file is uploaded then the name changes so it would be refreshed. This really can be cached forever.

zqian: Got it. Good to have those info in the comment.

jonespm:
I think this is good to merge, just waiting to see if anyone had any feedback on the comments.

zqian:
Comments look good to me. Please go ahead and merge. Thanks!

backend/urls.py
ssciolla: Did you find a doc for how to set `extra_headers` with this library? Can you link to that if so? Here or in the source code, I'm find either way.

You could also do something like `(365 * 24 * 60 * 60)` for increased readability on the age.

ssciolla (approved):
Didn't test, but seems fine. Couple comments.

backend/urls.py
jonespm: It was on an [issue filed against this library.](https://github.com/victor-o-silva/db_file_storage/issues/103)

Issue #219, opened 2022-06-23
Define only scopes needed by the app on LTI configuration


jonespm:
This application requests many scopes that it doesn't actually need or use and it should be reduced so only the required scopes are included. This is similar to reducing the scopes for the API in #202 .

https://github.com/tl-its-umich-edu/canvas-app-explorer/blob/main/templates/lti_config_template.json

jonespm:
Will be resolved by #246

PR #247, opened 2022-07-19
Limit access to the tool to course staff (#219, #246)

### LINKED ISSUE #246: Limit tool launches and tool visibility to users with staff roles
Students are currently not blocked in any way from using CAE, so the tool still launches, with some less than ideal behavior: a smaller subset of tools are listed and the add/remove buttons throw errors (as one would expect with OAuth implemented). Until we spend more time on handling the student role (#29), I think the most straightforward solution would be to block launches for any roles other instructor, admin, and sub-account admin. We may also want to limit tool visibility to “admins” for now.

TEST PLAN:

- Log in as a Canvas Admin, SubAccount Admin or course instructor, go to the course Settings->Navigation, and enable the Canvas App Explorer tool
- Verify that you can see the CAE tool on the left navigation, and successfully launch it
- Become student or log in as a student to this course, verify that the CAE tool is hidden for you in the left tool navigation.

ssciolla:
The PR aims to resolve #246.

### DIFF
--- a/backend/canvas_app_explorer/lti1p3.py
+++ b/backend/canvas_app_explorer/lti1p3.py
@@ -1,12 +1,13 @@
-import logging, random, string
+import logging, random, string, urllib.parse
 from collections import namedtuple
 from datetime import datetime
+from enum import Enum
 from typing import Any, Dict, Union
-import urllib.parse
 
 from django.conf import settings
 from django.contrib.auth import login as django_login
 from django.contrib.auth.models import User
+from django.core.exceptions import PermissionDenied
 from django.http import HttpRequest, HttpResponse, JsonResponse
 from django.shortcuts import redirect
 from django.urls import reverse
@@ -16,12 +17,24 @@ from django.views.decorators.http import require_POST
 from pylti1p3.contrib.django import DjangoOIDCLogin, DjangoMessageLaunch, \
     DjangoCacheDataStorage, DjangoDbToolConf
 
+
 logger = logging.getLogger(__name__)
+
+
+class CanvasRole(Enum):
+    ACCOUNT_ADMIN = 'Account Admin'
+    SUB_ACCOUNT_ADMIN = 'Sub-Account Admin'
+    TEACHER = 'TeacherEnrollment'
+    STUDENT = 'StudentEnrollment'
+
+
+STAFF_COURSE_ROLES = [CanvasRole.ACCOUNT_ADMIN, CanvasRole.SUB_ACCOUNT_ADMIN, CanvasRole.TEACHER]
 
 COURSE_MEMBERSHIP = 'http://purl.imsglobal.org/vocab/lis/v2/membership'
 DUMMY_CACHE = 'DummyCache'
 
 TOOL_CONF = DjangoDbToolConf()
+
 
 # do not require deployment ids if LTI_CONFIG_DISABLE_DEPLOYMENT_ID_VALIDATION is true
 class ExtendedDjangoMessageLaunch(DjangoMessageLaunch):
@@ -30,6 +43,8 @@ class ExtendedDjangoMessageLaunch(DjangoMessageLaunch):
             return self
 
         return super().validate_deployment()
+
+
 def lti_error(error_message: Any) -> JsonResponse:
     """
     Log an error message and return a JSON response with HTTP status 500.
@@ -43,6 +58,7 @@ def lti_error(error_message: Any) -> JsonResponse:
 
 def generate_jwks() -> Dict[str, list]:
     return TOOL_CONF.get_jwks()
+
 
 def get_jwks(_: Any) -> JsonResponse:
     """
@@ -100,7 +116,8 @@ def get_cache_config():
     cache_lifetime = cache_ttl if cache_ttl else 7200
     return CacheConfig(is_dummy_cache, launch_data_storage, cache_lifetime)
 
-def create_user_in_django(request, message_launch):
+
+def create_user_in_django(request: HttpRequest, message_launch: ExtendedDjangoMessageLaunch):
     launch_data = message_launch.get_launch_data()
     logger.debug(f'lti launch data {launch_data}')
     custom_params = launch_data['https://purl.imsglobal.org/spec/lti/claim/custom']
@@ -113,6 +130,14 @@ def create_user_in_django(request, message_launch):
     roles = launch_data['https://purl.imsglobal.org/spec/lti/claim/roles']
     username = custom_params['user_username']
     course_id = custom_params['canvas_course_id']
+    course_roles = custom_params['canvas_course_roles'].split(',')
+
+    staff_course_roles = [course_role for course_role in course_roles if course_role in STAFF_COURSE_ROLES]
+    user_is_course_staff = len(staff_course_roles) > 0
+    if not user_is_course_staff:
+        logger.info(f'User {username} does not have a staff role.')
+        raise PermissionDenied()
+
     if 'email' not in launch_data.keys():
         logger.info('Possibility that LTI launch by Instructor/admin becoming Canvas Test Student')
         error_message = 'Student view is not available for Canvas App Explorer.'
@@ -144,6 +169,7 @@ def create_user_in_django(request, message_launch):
     else:
         raise Exception('The canvas_course_id custom LTI variable must be configured.')
 
+
 @csrf_exempt
 def login(request):
     target_link_uri = request.POST.get('target_link_uri', request.GET.get('target_link_uri'))
@@ -153,6 +179,7 @@ def login(request):
     CacheConfig = get_cache_config()
     oidc_login = DjangoOIDCLogin(request, TOOL_CONF, launch_data_storage=CacheConfig.launch_data_storage)
     return oidc_login.enable_check_cookies().redirect(target_link_uri)
+
 
 @require_POST
 @csrf_exempt
--- a/backend/canvas_app_explorer/lti1p3.py
+++ b/backend/canvas_app_explorer/lti1p3.py
@@ -1,7 +1,6 @@
 import logging, random, string, urllib.parse
 from collections import namedtuple
 from datetime import datetime
-from enum import Enum
 from typing import Any, Dict, Union
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


backend/canvas_app_explorer/lti1p3.py
```diff
@@ -16,20 +17,34 @@
...
 
+
+class CanvasRole(Enum):
+    ACCOUNT_ADMIN = 'Account Admin'
+    SUB_ACCOUNT_ADMIN = 'Sub-Account Admin'
+    TEACHER = 'TeacherEnrollment'
+    STUDENT = 'StudentEnrollment'
+
+
+STAFF_COURSE_ROLES = [CanvasRole.ACCOUNT_ADMIN, CanvasRole.SUB_ACCOUNT_ADMIN, CanvasRole.TEACHER]
```
ssciolla, line 31: This could be moved to a different file.

backend/canvas_app_explorer/lti1p3.py
```diff
@@ -113,6 +130,14 @@ def create_user_in_django(request, message_launch):
     roles = launch_data['https://purl.imsglobal.org/spec/lti/claim/roles']
     username = custom_params['user_username']
     course_id = custom_params['canvas_course_id']
+    course_roles = custom_params['canvas_course_roles'].split(',')
+
+    staff_course_roles = [course_role for course_role in course_roles if course_role in STAFF_COURSE_ROLES]
+    user_is_course_staff = len(staff_course_roles) > 0
+    if not user_is_course_staff:
+        logger.info(f'User {username} does not have a staff role.')
+        raise PermissionDenied()
```
ssciolla, line 139: Still thinking about the optimal log and error messaging and implementation here.

backend/canvas_app_explorer/lti1p3.py
```diff
@@ -16,20 +17,34 @@
...
 
+
 logger = logging.getLogger(__name__)
 
+
+class CanvasRole(Enum):
+    ACCOUNT_ADMIN = 'Account Admin'
+    SUB_ACCOUNT_ADMIN = 'Sub-Account Admin'
+    TEACHER = 'TeacherEnrollment'
+    STUDENT = 'StudentEnrollment'
```
zqian, line 28: Maybe no need to define the student role if student is not allowed to use the tool

backend/canvas_app_explorer/lti1p3.py
zqian: maybe the log level should be WARN or ERROR?

backend/canvas_app_explorer/lti1p3.py
jonespm: Also these roles are already defined in the library. I'm not sure we need to redefine them.

https://github.com/dmitry-viskov/pylti1.3#check-users-role-after-lti-launch

We should just be able to check this with

`message_launch.check_staff_access() or message_launch.check_admin_access()`

It doesn't look like there's anything specific about sub account admin but we don't need that yet role yet either.

ssciolla: This is also what CCM does.

jonespm: I see, so there has been some discussion lately about [supporting sending a masquerade value](https://github.com/IMSGlobal/LTI-central/issues/157) over in the LTI launch. Right now I think you'd have to send that as a custom parameter if you cared about it. (Like we might if/when we supported students)

I'd believe this is an edge case we don't have to be concerned with now. If they're masquerading as a student the instructor won't have access to the tool in the navigation after the change here. And it's not going to give them a more limited scoped token if they did access it directly. 

I'm also not sure why it wouldn't be working.

Launching as a Instructor/Admin I see
```
  "https://purl.imsglobal.org/spec/lti/claim/roles": [
    "http://purl.imsglobal.org/vocab/lis/v2/institution/person#Administrator",
    "http://purl.imsglobal.org/vocab/lis/v2/institution/person#Instructor",
    "http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student",
    "http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor",
    "http://purl.imsglobal.org/vocab/lis/v2/system/person#User"
  ],
```
and
```
  "https://purl.imsglobal.org/spec/lti/claim/custom": {
    "roles": "TeacherEnrollment,Account Admin",
```

Doing Student View it has
```
  "https://purl.imsglobal.org/spec/lti/claim/roles": [
    "http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student",
    "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner",
    "http://purl.imsglobal.org/vocab/lis/v2/system/person#User",
    "http://purl.imsglobal.org/vocab/lti/system/person#TestUser"
  ],
```
and

```
  "https://purl.imsglobal.org/spec/lti/claim/custom": {
    "roles": "StudentEnrollment",
```

jonespm: I think it's fine since we're a little time crunched to go with this approach you have here for now. But we should file an issue for looking to support the standard LTI Roles when/if we support more roles and features in the future.

jonespm: Ah, I misread this as Student View verses Masquerade. This action in our instance is limited to someone that's an admin (or account admin).

It looks like it still sends `Instructor` as an institution role and the pylti1.3 [includes this as a common role](https://github.com/dmitry-viskov/pylti1.3/blob/master/pylti1p3/roles.py) for a Teacher. 

I'd think if you wanted a non masqueraded common role here you'd have to have only a context role. Though I don't think this "View As" edge case is going to really be a problem. They still do technically have the role as an Instructor and a Student in the same course. 

```
class TeacherRoleContextOnly(AbstractRole):
    _context_roles = ('Instructor', 'Administrator')  # type: tuple
```

ssciolla: I tried something like this class, and it's not really working as I expected. Code below:

```
class StaffRoleContextOnly(AbstractRole):
    _institution_roles = ('Administrator',)
    _context_roles = ('Instructor', 'Administrator')
```

```
    logger.info(roles)
    staff_role_obj = StaffRoleContextOnly(message_launch._get_jwt_body())
    for role in roles:
        logger.info(staff_role_obj.parse_role_str(role))

    user_is_staff_in_course = StaffRoleContextOnly(message_launch._get_jwt_body()).check()
    
    if not user_is_staff_in_course:
        logger.warn(f'User {username} does not have a staff role.')
        raise PermissionDenied()
```

This will work for the admin and student cases, but I couldn't launch when I was masquerading as an instructor. My initial impression is that the context roles generated by Canvas are malformed, or there's a bug in this library, but not sure exactly. When I parse the strings Canvas generates as the library does, I get this:
```
[INFO] [lti1p3.py:145] ['http://purl.imsglobal.org/vocab/lis/v2/institution/person#Instructor', 'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student', 'http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor', 'http://purl.imsglobal.org/vocab/lis/v2/system/person#User']
canvas_app_explorer        | [2022-07-20 17:32:33 +0000] canvas_app_explorer        | [2022-07-20 17:32:33 +0000] [INFO] [lti1p3.py:148] ('Student', 'institution')
[2022-07-20 17:32:33 +0000] [INFO] [lti1p3.py:148] ('Instructor', None)
[2022-07-20 17:32:33 +0000] [INFO] [lti1p3.py:148] ('User', 'system')

Comment (ssciolla) [reply backend/canvas_app_explorer/lti1p3.py:28]: @jonespm, it looks there may be an issue with the `pylti1.3` source code:
https://github.com/dmitry-viskov/pylti1.3/blob/7474685aecc35d528b944caf05b28ea66ac6d204/pylti1p3/roles.py#L44-L57

http://www.imsglobal.org/spec/lti/v1p3/#lis-vocabulary-for-context-roles

The `membership` roles just don't have the same structure as the institution or system roles; there is no slash, so the splits don't result in `('Instructor`, 'membership')` as expected.

I think I'm going to stick with the custom variable roles approach for now, but sure, we can create an issue to do this. It'd be interesting to look into this for CCM as well, but I don't think `ltijs` has an equivalent role class. Then again, these tools are pretty reliant on Canvas, and we're not getting much from the LMS-agnostic aspect of the LTI/LIS roles.

File: backend/canvas_app_explorer/lti1p3.py (hunk 2 above)
Comment (ssciolla) [reply backend/canvas_app_explorer/lti1p3.py]: Changed, see https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/247/commits/76a6c58050af25aba64d2af7ecd4b47a92844902

File: backend/canvas_app_explorer/lti1p3.py (hunk 3 above)
Comment (ssciolla) [reply backend/canvas_app_explorer/lti1p3.py]: @zqian , I removed the Student role. We may need it in the future. https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/247/commits/76a6c58050af25aba64d2af7ecd4b47a92844902

File: backend/canvas_app_explorer/lti1p3.py (hunk 1 above)
Comment (ssciolla) [reply backend/canvas_app_explorer/lti1p3.py]: Done, https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/247/commits/ff39b8e201d5628f8f544f7f7ed44e13d022e347

File: templates/lti_config_template.json [hunk 4]
```diff
@@ -1,21 +1,6 @@
 {
   "title": "Canvas App Explorer-%(timestamp)s",
-  "scopes": [
```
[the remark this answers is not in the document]
ssciolla: Issue #219? Guess we can close that too?

backend/canvas_app_explorer/lti1p3.py
jonespm: I'm fine with making a new issue and looking into it. But I'm also not convinced the "Act As" is really a true use case either. I think we'll have to pass the `Canvas.masqueradingUser.id` as a custom parameter until a better solution is worked out by IMS if we really cared about this use case. 

It does say when you "Act As" that 

> "You will be able to take any action as if you were this user, and from other users' points of views, it will be as if this user performed them. However, audit logs record that you were the one who performed the actions on behalf of this user."

So you still are the kind of both users when doing this. 

When the author committed this he said that pull requests were welcome against this if it's not splitting it correctly. I was thinking it probably could use some unit tests but wasn't about to prioritize that right now.

jonespm (approved):

templates/lti_config_template.json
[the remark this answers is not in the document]
zqian: +1!

zqian (approved):

backend/canvas_app_explorer/lti1p3.py
ssciolla: Yeah, probably a pull request with unit tests would be a good addition if we wanted to rely on this.

I'm still not quite sure what you mean by "I'm also not convinced the "Act As" is really a true use case either". The role identifiers I listed above for a masqueraded instructor don't show admin roles, and at least as far as the context ('`membership`, what's at issue here) there's no indication someone is masquerading as the user.

ssciolla: I will create a new issue.

PR #250, opened 2022-07-20
Announce number of filter results using `aria-live` (#248)

### LINKED ISSUE #248: Accessibility issue with Search field
Brandon Werner reported:

"When typing into the filed, screen readers don't speak that content is dynamically updating, and screen readers may expect, and be confused why they cannot find a search button. 

We recommend setting an aria-live region="polite" on the page containing the number of search results. Aria-live regions are used to tell assistive technology like screen readers about important content changes on a web page. As the user types, the number of search results should be changed in this region which will cause the screen reader to voice them. For maximum compatibility, this text should exist when the page loads. For more information about these regions, see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live."

TEST PLAN:

- Go to CAE tool and enter some text into the Search field
- If there is a match, info text "x out of y tools displayed" will be displayed in the middle. The x is the number of tools matched the search text, and the y is the total count of LTI tools 
- If there is no match, "No matching results" will be displayed instead.

ssciolla:
This PR aims to resolve #248.


zqian (approved):
works as expected.

jonespm:
It looks like it works, but it feels like it adds more whitespace to the top. 

Is this something that a sighted user cares about or something we should just hide and only be announced for a screen reader? Is it something that would be better placed somewhere else on the page? (These are more of UX questions) 

On Canvas App Explorer it says "Showing X Apps" but it's on a row with other sorting actions, rather than alone on its own row.

I guess this is something we could improve after release if there's no good fix ideas.

ssciolla:
> It looks like it works, but it feels like it adds more whitespace to the top.

Agreed. I think it looks OK, but not amazing.

> Is this something that a sighted user cares about or something we should just hide and only be announced for a screen reader? Is it something that would be better placed somewhere else on the page? (These are more of UX questions)

We discussed this in the watercooler. Seems like it does give sighted users a little more information (like how many tools are on the page without scrolling), and it seems common practice in other web applications.

Where else would you like it be? Putting it underneath the cards erases the sighted benefit I already mentioned. Spacing could be modified around, or I could try left or right alignment, though we haven't done that much yet. I will ask for the UX team's opinion.

ssciolla:
I'll merge now, and ask for the UX team feedback.

jonespm:
I'm not super sure. That line and the header itself take up almost 10% of the top area and it's feels more noticeable on mobile. 

I don't think moving it is a blocker but just want more space for the cards. 

Maybe we could make it an [Input Adornment](https://mui.com/material-ui/react-text-field/#input-adornments) on top or bottom of the search box but still have aria announce the change.

[This example ](https://mui.com/material-ui/react-text-field/#useformcontrol) shows when it changes but I didn't test if it was accessible.

PR #253, opened 2022-07-21
Perform dependency update and review (#252)

### LINKED ISSUE #252: Do pre-release dependency review and update
Prior to the initial release, it makes sense to check dependency versions for Python and Node packages and pick up patch or minor versions if no modifications to source code are needed.

ssciolla:
The PR aims to resolve #252.


requirements.txt
```diff
@@ -1,31 +1,27 @@
...
+markdown==3.4.1 # Markdown support for the browsable API
 
-#LTI 
-PyLTI1p3==1.9.1
-
-# Not in pypi https://github.com/Harvard-University-iCommons/django-canvas-oauth
-https://github.com/Harvard-University-iCommons/django-canvas-oauth/archive/v1.1.0.tar.gz
-
-# For serving static files
-whitenoise==5.3.0
```
ssciolla, line 26: Looks like there is a major release for `whitenoise`. Seems like it'd be low-risk, since we don't actually use the API very thoroughly and the Django version is supported. Thoughts? https://whitenoise.evans.io/en/stable/changelog.html

requirements.txt
```diff
@@ -1,31 +1,27 @@
-gunicorn
-
-Django>=3.2,<3.3
-django-webpack-loader==1.5.0
-mysqlclient
-django-mysql
-
-django-db-file-storage # Support for storage in the database
-Pillow # Needed for images
```
ssciolla, line 9: @jonespm, I removed `Pillow`, since things seemed to work without it. Was this one of the libraries you were thinking about that was unused? Others? Most everything else seemed needed.

ssciolla: Nevermind, seeing errors now, I'll add back `Pillow`. Must have seen some caching behavior.

frontend/package.json
```diff
@@ -9,33 +9,33 @@
...
-    "webpack": "^5.72.1",
-    "webpack-bundle-tracker": "1.5.0",
-    "webpack-cli": "^4.9.2",
+    "eslint": "^8.20.0",
+    "eslint-plugin-react": "^7.30.1",
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "mini-css-extract-plugin": "^2.6.1",
+    "ts-loader": "^9.3.1",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.7.4",
```
ssciolla, line 35: I upgraded TS, which is slightly riskier update, but I checked the release notes to ensure none of the breaking changes would affect us: https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#breaking-changes

requirements.txt
jonespm: Yeah pillow is needed for `models.ImageField`.

requirements.txt
jonespm: I don't see anything in the changelog that should break, just different versions supported. Seems fine if it works. You'll know immediately if it doesn't, though I *think* it's only used in the Openshift build.

ssciolla: Done, we'll have to test once it's built in OpenShift.

frontend/package.json
jonespm: Okay, sounds good!

jonespm (approved):
Didn't test out all the frontend but looks good to me.

Issue #256, opened 2022-07-22
create migration file for the ltitool short_description field change introduced in #179


zqian:

zqian:
close since this is a duplicate of #255

ssciolla:
This can stay open if you prefer to have a new issue rather than re-opening #179.

PR #255, opened 2022-07-22
Add missing migration for added validator (#256)

ssciolla:
The PR resolves #256.

PR #257, opened 2022-07-22
#256 added migration file for ltitool model change

zqian:


zqian:
Close since this is duplicate of #255

PR #259, opened 2022-07-25
Add `django-watchman` status URL

### LINKED ISSUE #258: Add `django-watchman` status check endpoint
Not sure if this is 100 percent required for the initial release, but it is something we use in our other Django applications. https://github.com/mwarkentin/django-watchman

ssciolla:
The PR aims to resolve #258.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -22,6 +22,11 @@
 # The URL for an external help resource; defaults to an empty string
 # HELP_URL=https://github.com/tl-its-umich-edu/canvas-app-explorer
 
+# (optional) The token for accessing the /watchman URL; leaving it blank means the route is unprotected
+# DJANGO_WATCHMAN_TOKENS=
+# (optional) The name of the header or parameter used for passing the above token
+# DJANGO_WATCHMAN_TOKEN_NAME=
+
 # A comma separated list of domains where you want to allow the application to be framed
 # CSP_FRAME_ANCESTORS=umich.instructure.com
 
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -203,7 +203,10 @@ LOGGING = {
             'level': 'WARNING',
             'handlers': ['console'],
         },
-
+        'watchman': {
+            'handlers': ['console'],
+            'level': 'DEBUG',
+        }
     },
     'root': {
         'level': os.getenv('ROOT_LOG_LEVEL', 'INFO'),
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -219,15 +219,10 @@ TINYMCE_DEFAULT_CONFIG = {
         "default_link_target": "_blank",
     }
 
-# Watchman settings
-WATCHMAN_CHECKS = (
-    'watchman.checks.caches',
-    'watchman.checks.databases'
-)
-if isinstance((env_watchman_tokens := os.getenv('DJANGO_WATCHMAN_TOKENS')), str):
-    WATCHMAN_TOKENS = env_watchman_tokens.split(',')
-    WATCHMAN_TOKEN_NAME = os.getenv('DJANGO_WATCHMAN_TOKEN_NAME', 'token')
-
+# Watchman settings (https://github.com/mwarkentin/django-watchman)
+WATCHMAN_TOKENS = os.getenv('DJANGO_WATCHMAN_TOKENS', None)
+WATCHMAN_TOKEN_NAME = os.getenv('DJANGO_WATCHMAN_TOKEN_NAME', 'token')
+WATCHMAN_CHECKS = ('watchman.checks.caches', 'watchman.checks.databases')
 
 # OAuth Settings, get these from the environment
 # TODO: Add some of these to the Django LTI model
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -224,7 +224,7 @@ TINYMCE_DEFAULT_CONFIG = {
 
 # Watchman settings (https://github.com/mwarkentin/django-watchman)
 WATCHMAN_TOKENS = os.getenv('DJANGO_WATCHMAN_TOKENS', None)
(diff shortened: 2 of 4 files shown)


backend/settings.py
```diff
@@ -218,6 +219,11 @@
         "default_link_target": "_blank",
     }
 
+# Watchman settings (https://github.com/mwarkentin/django-watchman)
+WATCHMAN_TOKENS = os.getenv('DJANGO_WATCHMAN_TOKENS', None)
+WATCHMAN_TOKEN_NAME = os.getenv('DJANGO_WATCHMAN_TOKEN_NAME', 'token')
+WATCHMAN_CHECKS = ('watchman.checks.caches', 'watchman.checks.databases')
```
ssciolla, line 225: This is defined in order to exclude the default `watchman.checks.storage`, which is failing due to our use of [`django-db-file-storage`](https://pypi.org/project/django-db-file-storage/). There may be a way to fix this, but it would require a custom `django-watchman` check. I'm also not sure how useful the storage check is if we are ultimately storing the file content in the database; maybe the database check is enough.

.env.sample
```diff
@@ -22,6 +22,11 @@
 # The URL for an external help resource; defaults to an empty string
 # HELP_URL=https://github.com/tl-its-umich-edu/canvas-app-explorer
 
+# (optional) The token for accessing the /watchman URL; leaving it blank means the route is unprotected
+# DJANGO_WATCHMAN_TOKENS=
+# (optional) The name of the header or parameter used for passing the above token
+# DJANGO_WATCHMAN_TOKEN_NAME=
```
pushyamig, line 28: I was successfully able to check the watchman endpoint receiving the DB, cache status using a query parameter via GET call from the browser.  When I removed the token from the ConfigsMap the endpoint became optional and it worked very well.

.env.sample
```diff
@@ -22,6 +22,11 @@
 # The URL for an external help resource; defaults to an empty string
 # HELP_URL=https://github.com/tl-its-umich-edu/canvas-app-explorer
 
+# (optional) The token for accessing the /watchman URL; leaving it blank means the route is unprotected
```
pushyamig, line 25: When I was trying to make a call with authorization header via curl or postman, I received a forbidden error.  This is how I made my query, not sure I did this right

`curl -X GET -H "Authorization:  token=\”xyh\”” https://dev-cae/watchman`

backend/settings.py
pushyamig: I think this is ok. We can add this `storage` later iteration if needed.

requirements.txt
```diff
@@ -7,6 +7,7 @@ django-csp==3.7 # For Content Security Policy
 django-db-file-storage==0.5.5 # Support for storage in the database
 django-mysql==4.7.0
 django-tinymce==3.4.0 # Rich text editor
+django-watchman==1.3
```
pushyamig, line 10: I saw from the [Docs](https://github.com/mwarkentin/django-watchman#logging) of this library that we can add the watchman to the Django logging configuration. But this PR doesn't account for that, but I do see some level of watchman logging in OpenShift pod.  

Why did you opt out of not adding the watchman logging config? Either way is fine, just wanted to understand your rationale

backend/settings.py
pushyamig, line 225: Curious question, what is it that you are caching in CAE? purpose of the caching in APP?

ssciolla: Good question, I believe it's just database caching, but not sure. See `DB_CACHE_CONFIGS` in `settings.py`.

pushyamig: Yeah. It is DB caching and MyLA too has it for storing the PyLTI library token info w.r.t launch. Like how you created a Browser cookie for LTI user as a JWT appending as a header in CCM. 

Now thinking back on the cache watchman status endpoint, probably that is not strictly needed since the cache is stored in DB. So a status check on DB is sufficient. But I might not know if there is any other way this app uses cache.

ssciolla: Yeah, I'm not sure. I don't think there's any harm in including this, so I'm inclined to keep it as is for now.

.env.sample
ssciolla: I got this to work. Looks like you would form it like:
```
curl -X GET -H "Authorization: {your custom token name} Token=\”xyh\”” https://dev-cae/watchman
```
The environment variable value goes after `Authorization: `. You have to use the `Token=` form after that.

requirements.txt
ssciolla: I didn't see this in MyLA, so I guess I wasn't too worried about it. I added it though at the `DEBUG` level. It's pretty noisy, there may be some template issues with `watchman`, but they don't seem to affect functionality.

backend/settings.py
```diff
@@ -218,6 +222,11 @@
         "default_link_target": "_blank",
     }
 
+# Watchman settings (https://github.com/mwarkentin/django-watchman)
+WATCHMAN_TOKENS = os.getenv('DJANGO_WATCHMAN_TOKENS', None)
+WATCHMAN_TOKEN_NAME = os.getenv('DJANGO_WATCHMAN_TOKEN_NAME', 'token')
```
pushyamig, line 227: Optional: I feel the `token` name is too generic. I would make it `cae_wm_token` or `wm_token`

.env.sample
pushyamig: I think this will work since AWS outage the app is not reachable.

pushyamig (approved):
I was able to test it and the approach with a protected status URL and make it optional working very well. 
What remains to understand is whether Nagios can make an HTTP call to a protected URL or not. Regardless using the watchman URL or not,  Nagios setup could be as basic as call the root URL and CAE project team will decide on that.

backend/settings.py
ssciolla: Okay, I'll change it.

PR #261, opened 2022-07-29
Add `role`, `aria-hidden`, and `focusable` values to `CheckCircleOutline` icon (#260)

### LINKED ISSUE #260: Add `aria-hidden=false` and `role='button'` to focusable icon in `ToolCard`
From the accessibility team:

Issue: The control that invokes the tooltip is an <svg> that admits focus but has an aria-hidden=“true” attribute. Normally <svg> and font icons that are decorative should have that attribute, but not when they are informational.

To Fix: Remove the aria-hidden attribute and might as well give it a role=“button” since that is what it is, really.

ssciolla:
This PR aims to resolve #260.

### DIFF
--- a/frontend/app/components/ToolCard.tsx
+++ b/frontend/app/components/ToolCard.tsx
@@ -137,7 +137,14 @@ export default function ToolCard (props: ToolCardProps) {
           {
             tool.navigation_enabled && (
               <Tooltip placement='top' title={TOOL_IN_MENU_TEXT}>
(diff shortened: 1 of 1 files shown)


pushyamig:
This is working as expected, I was able to test this from CAE Dev instance build from this branch. The circle icon is focusable and screen reader is able to identify the icon as button.

pushyamig (approved):

PR #264, opened 2022-08-02
Added metadata field to lti tool model, visible to admin only (#10)

### LINKED ISSUE #10: Have a place for admins to store internal metadata (like notes and tool configuration information)

roujiawang:
Resolved #10


backend/canvas_app_explorer/models.py
```diff
@@ -39,6 +39,7 @@ class LtiTool(models.Model):
     privacy_agreement = HTMLField()
     support_resources = HTMLField()
     canvas_placement = models.ManyToManyField(CanvasPlacement)
+    metadata = HTMLField(null=True)
```
ssciolla, line 42: @jonespm, how do we feel about the name `metadata`, as opposed to just `notes` or `internal_notes`?

ssciolla: I could see adding `help_text` here to indicate this isn't shown to users, especially if we don't change the name to internal. Something like this:
```suggestion
    metadata = HTMLField(null=True, help_text='Field for internal notes (not shown to users)')
```

ssciolla (commented):
This is a functional and seems like it would be useful, at least during the content creation or editing phase for pasting snippets or tracking progress on review, etc. I just wonder if `metadata` is the appropriate name. I'd like to hear @jonespm's or @ziegler1960's opinion.

backend/canvas_app_explorer/models.py
jonespm: Yeah, I think internal_notes sounds like a better field name.

ziegler1960: Reading thru this request it's not entirely clear to me why we need this field and what will be entered.  Nonetheless, I agree that Internal notes is better than 'metadata'

ssciolla (approved):
Nice work! With one approval, you can merge this now. Click "Squash and merge" at the bottom when you're ready.

PR #265, opened 2022-08-02
Fix usage of `django-watchman` (#263)

### LINKED ISSUE #263: Change `watchman` URL mount point to `status`, fix logging
To keep consistent with other applications, we should mount the `watchman` application URLs at `status` instead of `watchman`. We may also want to add a `bare_status` endpoint.

We should also change the log level in the logger so it only logs messages with level `INFO` or `WARNING` or higher. It could be configurable also.

ssciolla:
The PR aims to resolve #263.

### DIFF
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -188,30 +188,17 @@ LOGGING = {
             'formatter': 'generic',
         },
     },
+    'root': {
+        'level': os.getenv('ROOT_LOG_LEVEL', 'INFO'),
+        'handlers': ['console']
+    },
     'loggers': {
         'django': {
             'handlers': ['console'],
             'propagate': False,
             'level': os.getenv('DJANGO_LOG_LEVEL', 'INFO'),
-        },
-        'rules': {
-            'handlers': ['console'],
-            'propagate': False,
-            'level': os.getenv('RULES_LOG_LEVEL', 'INFO'),
-        },
-        '': {
-            'level': 'WARNING',
-            'handlers': ['console'],
-        },
-        'watchman': {
-            'handlers': ['console'],
-            'level': os.getenv('DJANGO_WATCHMAN_LOG_LEVEL', 'INFO'),
         }
-    },
-    'root': {
-        'level': os.getenv('ROOT_LOG_LEVEL', 'INFO'),
-        'handlers': ['console']
-    },
+    }
 }
 
 TINYMCE_DEFAULT_CONFIG = {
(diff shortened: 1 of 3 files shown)


backend/settings.py
```diff
@@ -205,7 +205,7 @@
         },
         'watchman': {
             'handlers': ['console'],
-            'level': 'DEBUG',
+            'level': os.getenv('DJANGO_WATCHMAN_LOG_LEVEL', 'INFO'),
```
jonespm, line 208: We probably don't need separately configurable log levels in here and could just use one setting for all loggers this is fine for now.

jonespm (approved):

backend/settings.py
ssciolla: Okay, so what's the ideal configuration here? I see Django and Rules have separate environment variables. I'm guessing they could all be the same, or maybe we just remove `level` and have them rely on the `root` logger, if it works like that?

ssciolla: Also, what is the purpose of the `''` logger?

jonespm: Yeah, if we just removed the individual loggers for each package it would use the root. I don't think this application is complex enough that we'd need to adjust levels for any individual packages. 

I _think_ we could just simplify this and remove the entire loggers block leaving a simple configuration more similar to the first example on [Logging](https://docs.djangoproject.com/en/4.0/topics/logging/#examples). (But leaving it configurable) 

The comment on 
https://lincolnloop.com/blog/django-logging-right-way/

Says that `''` is the root logger. I wonder if this is something old before `'root'` was explicitly added? [This post](https://www.reddit.com/r/django/comments/kw835z/django_logging_configuration_difference_between/) doesn't really clear anything up. :) I'd probably switch to the simple example on djangoproject.com that has `'root'`without the empty logger.

ssciolla: How is this? https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/265/commits/0b624877c7511fbff972e6457196a459e4882e7a I think there might be some benefit to being able to turn the knob on Django logging separately, since the SQL emitted at the DEBUG level can be really noisy.

ssciolla: Hi @jonespm, I re-requested a review here because I wanted your feedback on the logging changes I made after you approved. If you can take a quick look, that would be great.

jonespm: Sure, this still looks good. I don't think we need every logger individually customizable here.

jonespm (approved):

Issue #203, opened 2022-06-16
text format consistency (text font, text size) for all text input fields


zqian:
CAE supports WYSIWYG text input for "short description", "long description", "privacy agreement", and "support resources". Those text field values will all be displayed in the "find more" screen. 

It is very likely for user to copy and paste text directly from external web pages, and hence those field values will contain different fonts and text sizes. Can we "sanitize" the html text to make a consistent style display in the "find more" page?

ssciolla:
"Because the admin editor (TinyMCE) currently allows anything, including copy and pasting styles, and the frontend presents this with dangerouslySetInnerHTML there was some desire to better restrict the tags and content that can be used both by the editor and displayed.

This could either be done in the backend when the content is delivered to the API or possibly in React. We have to determine what things we want to whitelist, like links, and possibly simple formatting like bold/italic/colors, and block everything else."

Comment by @jonespm in the duplicate issue (#217)

ziegler1960:
This fix has been verified -  after entering text with a variety of formatting into the TinyMCE and Saving, the Short Description output on the frontend ALP card displays a common font type and size, while preserving other formats from the original HTML such as inline links, bolding, and italics.

Test Course: https://umich.beta.instructure.com/courses/587202/external_tools/48681

- [ ] From the CAE tool on the test course, click Admin link at the upper right  to access the Admin UI
- [ ] Select 'Lti tools' to see the list of cards
- [ ] Click on a tool: Active Learning Platform
- [ ] Copy a selection of text with a variety of formats (font size and type, bold, italic, web link) from a web page (https://www.yahoo.com/entertainment/rolling-stones-announce-grrr-live-140053553.html) and paste into the short description field for the tool.
- [ ] Click Save - LTI tool list loads
- [ ] Click View Site link to return to the CAE tool on the course
- [ ] Verify that the Short Description output on the frontend ALP card has a common font type and size, while preserving other formats from the original HTML such as inline links, bolding, and italics.
- [ ] Return to Admin UI -> Lti tools ->  click on the ACC listing
- [ ] Restore the original text to the ALP Short Description field ("Active Learning Platform facilitates an engaging and active learning space for lectures.") and Save
- [ ] Return to the course and verify that the ALP Short Description is displaying the original text

PR #266, opened 2022-08-03
TinyMCE Default Config Modification (#203)

roujiawang:
Modified TinyMCE default config in settings to block certain html element and style tags when admin pastes texts into LTI tool card fields.

### DIFF
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -207,6 +207,9 @@ TINYMCE_DEFAULT_CONFIG = {
         "toolbar": "formatselect | bold italic backcolor | "
             "bullist numlist outdent indent | removeformat | code | help",
         "default_link_target": "_blank",
+        "invalid_elements": "span, h1, h2, h3, h4, h5, h6",
+        "invalid_styles": "color, font-family, background-color"
+
     }
 
 # Watchman settings (https://github.com/mwarkentin/django-watchman)
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -202,38 +202,47 @@ LOGGING = {
 }
 
 TINYMCE_DEFAULT_CONFIG = {
-        'menubar': True,
-        "plugins": "advlist,autolink,code,lists,link,anchor,insertdatetime,media,table,paste,wordcount",
-        "toolbar": "bold italic backcolor | "
-            "bullist numlist outdent indent | removeformat | code | help",
-        "default_link_target": "_blank",
-        "invalid_elements": "span, h1, h2, h3, h4, h5, h6",
-        "invalid_styles": "color, font-family, background-color",
+    'menubar': True,
+    "plugins": "advlist,autolink,code,lists,link,anchor,insertdatetime,media,table,paste,wordcount",
+    "toolbar": "bold italic backcolor | "
+        "bullist numlist outdent indent | removeformat | code | help",
+    "default_link_target": "_blank",
+    "invalid_elements": "span, h1, h2, h3, h4, h5, h6",
+    "invalid_styles": "color, font-family, background-color",
 
-        "style_formats": [
-            { "title": 'Inline', "items": [
-            { "title": 'Bold', "format": 'bold' },
-            { "title": 'Italic', "format": 'italic' },
-            { "title": 'Underline', "format": 'underline' },
-            { "title": 'Strikethrough', "format": 'strikethrough' },
-            { "title": 'Superscript', "format": 'superscript' },
-            { "title": 'Subscript', "format": 'subscript' },
-            { "title": 'Code', "format": 'code' }
-            ]},
-            { "title": 'Blocks',  "items": [
-            { "title": 'Paragraph', "format": 'p' },
-            { "title": 'Blockquote', "format": 'blockquote' },
-            { "title": 'Div', "format": 'div' },
-            { "title": 'Pre', "format": 'pre' }
-            ]},
-            { "title": 'Align',  "items": [
-            { "title": 'Left', "format": 'alignleft' },
-            { "title": 'Center', "format": 'aligncenter' },
-            { "title": 'Right', "format": 'alignright' },
-            { "title": 'Justify', "format": 'alignjustify' }
-            ]}
-            ],
-            "block_formats": "Paragraph=p",
+    "style_formats": [
+        { 
+            "title": 'Inline', 
+            "items": [
+                { "title": 'Bold', "format": 'bold' },
+                { "title": 'Italic', "format": 'italic' },
+                { "title": 'Underline', "format": 'underline' },
+                { "title": 'Strikethrough', "format": 'strikethrough' },
+                { "title": 'Superscript', "format": 'superscript' },
+                { "title": 'Subscript', "format": 'subscript' },
+                { "title": 'Code', "format": 'code' }
+            ]
+        },
+        { 
+            "title": 'Blocks',  
+            "items": [
+                { "title": 'Paragraph', "format": 'p' },
+                { "title": 'Blockquote', "format": 'blockquote' },
+                { "title": 'Div', "format": 'div' },
+                { "title": 'Pre', "format": 'pre' }
+            ]
+        },
+        { 
+            "title": 'Align',  
+            "items": [
+                { "title": 'Left', "format": 'alignleft' },
+                { "title": 'Center', "format": 'aligncenter' },
+                { "title": 'Right', "format": 'alignright' },
+                { "title": 'Justify', "format": 'alignjustify' }
+            ]
+        }
+        ],
+        "block_formats": "Paragraph=p",
     }
 
 # Watchman settings (https://github.com/mwarkentin/django-watchman)
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -208,7 +208,7 @@ TINYMCE_DEFAULT_CONFIG = {
         "bullist numlist outdent indent | removeformat | code | help",
     "default_link_target": "_blank",
     "invalid_elements": "h1, h2, h3, h4, h5, h6",
-    "invalid_styles": "color, font-family, background-color",
+    "invalid_styles": "color, font-family, background-color, line-height",
 
     "style_formats": [
         { 
@@ -243,6 +243,7 @@ TINYMCE_DEFAULT_CONFIG = {
         }
     ],
     "block_formats": "Paragraph=p",
+    "removed_menuitems": "fontformats fontsizes lineheight forecolor backcolor",
 }
 
 # Watchman settings (https://github.com/mwarkentin/django-watchman)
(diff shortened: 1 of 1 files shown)


backend/settings.py
```diff
@@ -220,6 +220,9 @@
         "toolbar": "formatselect | bold italic backcolor | "
             "bullist numlist outdent indent | removeformat | code | help",
         "default_link_target": "_blank",
+        "invalid_elements": "span, h1, h2, h3, h4, h5, h6",
+        "invalid_styles": "color, font-family, background-color"
+
```
ssciolla, line 225:
```suggestion
```

backend/settings.py
```diff
@@ -220,6 +220,9 @@
         "toolbar": "formatselect | bold italic backcolor | "
             "bullist numlist outdent indent | removeformat | code | help",
         "default_link_target": "_blank",
+        "invalid_elements": "span, h1, h2, h3, h4, h5, h6",
+        "invalid_styles": "color, font-family, background-color"
```
ssciolla, line 224: This seems to be working pretty well. I think we could allow `span`, there might be a use for it.

I'd probably want to check this with @zqian and Jeff next week to make sure they're on board with the change.

jonespm: Yeah, I'd also like to get more examples on what is really worth removing, most things are defined as classes rather than inline styles. 

It's nice that it's configurable here though. Most things on the web are probably defined as a div. 

I think if we removed these we'd also want to remove these options from the "Format" menu dropdown too, which would probably require [fully configuring the menu. ](https://www.tiny.cloud/docs/configure/editor-appearance/#menu)

ssciolla: @roujiawang, can you look into restricting the available element types from the menus, specifically headers?

ssciolla: If we can come up with a few test cases as well, that can be stored somewhere, that would be great.

jonespm: We may need to restrict the class attribute or only allow some classes as well. We'd have to look at the test cases and see what they include.

roujiawang: The settings for text box toolbar menu can be changed by overriding certain default features inside the braces of TINYMCE_DEFAULT_CONFIG in settings.py. For instance, adding _"block_formats": "",_ would clear out font size options in the "Paragraph" section of the toolbar menu, making admin editors unable to set font size h1~h6.
I could launch a new PR about this.

roujiawang: There may be gentler ways to disable certain html elements, like the "Using remove" example in the middle of https://www.tiny.cloud/docs/configure/content-formatting/

ssciolla: I think the idea was we'd like you to disable these options as part of this PR. I think the first approach sounds more promising. "using remove" doesn't seem to actually disable the elements, just modify the behavior of the "Remove formatting" button.

backend/settings.py
```diff
@@ -217,9 +217,36 @@
...
+            { "title": 'Pre', "format": 'pre' }
+            ]},
+            { "title": 'Align',  "items": [
+            { "title": 'Left', "format": 'alignleft' },
+            { "title": 'Center', "format": 'aligncenter' },
+            { "title": 'Right', "format": 'alignright' },
+            { "title": 'Justify', "format": 'alignjustify' }
+            ]}
+            ],
+            "block_formats": "Paragraph=p",
```
ssciolla, line 249: I'll test this today. I'd like to see this spacing normalized so it's easier to read. I'll take a stab at it.

ssciolla: So I'd like to see some formatting changes here. First, all this should be unindented once (it has two now). Then, it'd be nice to space the objects and arrays like below for better readability.
```
    'menubar': True,
    "plugins": "advlist,autolink,code,lists,link,anchor,insertdatetime,media,table,paste,wordcount",
    "toolbar": "bold italic backcolor | "
        "bullist numlist outdent indent | removeformat | code | help",
    "default_link_target": "_blank",
    "invalid_elements": "span, h1, h2, h3, h4, h5, h6",
    "invalid_styles": "color, font-family, background-color",
    "style_formats": [
        {
            "title": 'Inline',
            "items": [
                {"title": 'Bold', "format": 'bold'},
                {"title": 'Italic', "format": 'italic'},
                {"title": 'Underline', "format": 'underline'},
                {"title": 'Strikethrough', "format": 'strikethrough'},
                {"title": 'Superscript', "format": 'superscript'},
                {"title": 'Subscript', "format": 'subscript'},
                {"title": 'Code', "format": 'code'}
            ]
        },
        {
            "title": 'Blocks', 
            "items": [
                {"title": 'Paragraph', "format": 'p'},
                {"title": 'Blockquote', "format": 'blockquote'},
                {"title": 'Div', "format": 'div'},
                {"title": 'Pre', "format": 'pre'}
            ]
        },
        {
            "title": 'Align',
            "items": [
                {"title": 'Left', "format": 'alignleft'},
                {"title": 'Center', "format": 'aligncenter'},
                {"title": 'Right', "format": 'alignright'},
                {"title": 'Justify', "format": 'alignjustify'}
            ]
        }
    ],
    "block_formats": "Paragraph=p",
```
I'd also like to see consistent use of single or double quotation marks, but we're not consistent elsewhere in `settings.py`, so it's not the biggest deal.

ssciolla: I was hoping there'd be a way to not have to specify all this and just tell it to omit the heading menu, but I haven't found that yet.

backend/settings.py
```diff
@@ -215,11 +215,47 @@
...
+            "items": [
+                { "title": 'Left', "format": 'alignleft' },
+                { "title": 'Center', "format": 'aligncenter' },
+                { "title": 'Right', "format": 'alignright' },
+                { "title": 'Justify', "format": 'alignjustify' }
+            ]
+        }
+        ],
+        "block_formats": "Paragraph=p",
     }
```
ssciolla, line 259: Looks like these lines still need to be unindented once.

backend/settings.py
```diff
@@ -215,12 +215,48 @@
...
-        "toolbar": "formatselect | bold italic backcolor | "
-            "bullist numlist outdent indent | removeformat | code | help",
-        "default_link_target": "_blank",
-    }
+    'menubar': True,
+    "plugins": "advlist,autolink,code,lists,link,anchor,insertdatetime,media,table,paste,wordcount",
+    "toolbar": "bold italic backcolor | "
+        "bullist numlist outdent indent | removeformat | code | help",
+    "default_link_target": "_blank",
+    "invalid_elements": "span, h1, h2, h3, h4, h5, h6",
```
ssciolla, line 223: I believe we discussed the inclusion of `span` here before; was there a reason why this needs to be invalid? It seems like the header tags are the main thing we're trying to avoid here.

backend/settings.py
```diff
@@ -215,12 +215,48 @@
...
+        { 
+            "title": 'Align',  
+            "items": [
+                { "title": 'Left', "format": 'alignleft' },
+                { "title": 'Center', "format": 'aligncenter' },
+                { "title": 'Right', "format": 'alignright' },
+                { "title": 'Justify', "format": 'alignjustify' }
+            ]
+        }
+        ],
```
ssciolla, line 257:
```suggestion
    ],
```

ssciolla (commented):
A couple more things here to discuss or address.

I would also still like you to provide a few test cases to ensure this is working as expected. Basically this means we'd like sample HTML to drop into the editor. The ones that immediately come to mind are:
1) HTML with a different font and font size
2) HTML with header tags

You can share this with us directly (email, Google Drive). Let me know if you have questions.

backend/settings.py
roujiawang: Like shown in the example [here](https://www.w3schools.com/tags/tag_span.asp), the `span` tag is often used to specify the `style` of a particular part of texts (indentation, color, etc.). Excluding `span` from pasted texts was to remove the customized text styles that we may not want to have in our cards.

backend/settings.py
roujiawang: Fixed it. Thanks for pointing it out!

roujiawang:
For the current effect of html text filter settings I have created [this gist](https://gist.github.com/roujiawang/38ea8c28e024a299f1a3e808eaf9947b) to record the results of my local tests.

backend/settings.py
ssciolla: I'm still a bit uncertain about this. As you say, `span` is often used to specify styles, but it's not exactly the root of the problem; if we wanted to, I think we could disallow more styles using `invalid_styles` as has already been done. I guess my preference is to do that: allow `span` and disallow any specific styles we don't want. @jonespm, what do you think?

backend/settings.py
```diff
@@ -202,12 +202,48 @@
...
-            "bullist numlist outdent indent | removeformat | code | help",
-        "default_link_target": "_blank",
-    }
+    'menubar': True,
+    "plugins": "advlist,autolink,code,lists,link,anchor,insertdatetime,media,table,paste,wordcount",
+    "toolbar": "bold italic backcolor | "
+        "bullist numlist outdent indent | removeformat | code | help",
+    "default_link_target": "_blank",
+    "invalid_elements": "h1, h2, h3, h4, h5, h6",
+    "invalid_styles": "color, font-family, background-color",
```
ssciolla, line 211: I think we may want to add `line-height` to this list. The indentation thing I'm not so worried about, but weird line spacing could interfere with easy it is to tell the sections apart.

backend/settings.py
```diff
@@ -202,12 +202,48 @@
...
+            "title": 'Align',  
+            "items": [
+                { "title": 'Left', "format": 'alignleft' },
+                { "title": 'Center', "format": 'aligncenter' },
+                { "title": 'Right', "format": 'alignright' },
+                { "title": 'Justify', "format": 'alignjustify' }
+            ]
+        }
+    ],
+    "block_formats": "Paragraph=p",
```
ssciolla, line 245:
```suggestion
    "block_formats": "Paragraph=p",
    "removed_menuitems": "fontformats fontsizes lineheight forecolor backcolor",
```

I think we we'll want to add this line. It seems to remove hide a lot of the settings we disallowed. Happy to show you how I found this if you'd like. I looked at this setting (https://www.tiny.cloud/docs/configure/editor-appearance/#removed_menuitems) and the menu setting (to determine item names) in the version 5 docs (what `django-tinymce` seems to be using).

roujiawang: Thanks for the information! Line added.

backend/settings.py
roujiawang: I have added ```line-height``` to ```invalid_styles```, and the line height attributes get filtered out when I pasted the following html text into a text box in a card:   
```
<p style="line-height:10cm">10cm line height</br>
how does this look?</p>
<p style="line-height:80%">80% normal line height</br>
how does this look?</p>
```

ssciolla (approved):
Looks good, thanks for being patient with me with the multiple reviews. We can merge now.

jonespm (approved):
Thanks for working though this with @roujiawang @ssciolla. The changes look fine to me.

PR #275, opened 2022-09-23
Modified paragraph class attributes to make paragraphs separated by l…

roujiawang:
…ine breaks


ssciolla (approved):
Nice work! Thanks again for identifying, document, and fixing this issue.

PR #270, opened 2022-08-17
Change session engine to use cache, simplify and improve LTI launch (#262)

ssciolla:

PR #277, opened 2022-10-04
Add `redis` service to improve LTI launch performance

### LINKED ISSUE #262: Validate the launch closer to the actual launch
I believe some issues with launch validation might be because we are waiting a bit to do any launch validation.

If we explicitly validate the launch immediately after the launch we shouldn't encounter issues with the validation not matching, and if so be able to handle it better in our code.

https://github.com/dmitry-viskov/pylti1.3#lti-message-launches

I'm not sure if we'd need to implement any [locking around this](https://stackoverflow.com/questions/18213619/sharing-a-lock-between-gunicorn-workers). We're doing more testing. I believe that if the testing is successful just having a validate closer to the launch should be sufficient. 

https://github.com/tl-its-umich-edu/canvas-app-explorer/blob/a7a8f86bbb651b61e47c49716988db430c0dbca0/backend/canvas_app_explorer/lti1p3.py#L182

ssciolla:
This PR aims to resolve #262. This PR may supersede #270.


jonespm (approved):

requirements.txt
```diff
@@ -11,6 +11,7 @@ django-watchman==1.3
 django-webpack-loader==3.1.1
 Pillow==11.0.0
 whitenoise==6.2.0 # For serving static files
+django-redis==5.2.0
```
jonespm, line 14:
```suggestion
django-redis==5.4.0
```

backend/canvas_app_explorer/lti1p3.py
```diff
@@ -12,19 +12,29 @@
...
+    DjangoCacheDataStorage, DjangoDbToolConf, DjangoMessageLaunch, DjangoOIDCLogin
+)
+from pylti1p3.exception import LtiException
 
 from .canvas_roles import STAFF_COURSE_ROLES
 
+
 logger = logging.getLogger(__name__)
 
+COURSE_MEMBERSHIP = 'http://purl.imsglobal.org/vocab/lis/v2/membership'
```
zqian, line 25: Nothing is changed for the COURSE_MEMBERSHIP value

backend/settings.py
```diff
@@ -154,12 +154,26 @@
...
+    #     'CACHE_OPTIONS': {'COMPRESS_MIN_LENGTH': 5000, 'COMPRESS_LEVEL': 6}
+    # },
+    {
+        'CACHE_TTL': 600,
+        'BACKEND': 'django_redis.cache.RedisCache',
+        'LOCATION': f"redis://{os.getenv('REDIS_HOST', 'redis')}:6379/1",
+        'CACHE_KEY_PREFIX': 'app_explorer',
+        'CACHE_OPTIONS': {
+            'CLIENT_CLASS': 'django_redis.client.DefaultClient',
+            'PASSWORD': os.getenv('REDIS_PASS')
```
zqian, line 173: Should there be some default password string assigned, in case the env value is missing?

zqian (approved):

backend/settings.py
jonespm: By default in the local environment there isn't a password set and it starts up without it. But it's also not exposed to the internet and just internal. I think it would just give an error if it has a password and it's not correct, which would need to be addressed.

backend/settings.py
```diff
@@ -154,12 +154,26 @@
...
 
-DB_CACHE_CONFIGS = os.getenv('DB_CACHE_CONFIGS',
-                           {'CACHE_TTL': 600, 'BACKEND': 'django_mysql.cache.MySQLCache',
-                            'LOCATION': 'canvas_app_explorer_cache',
-                            'CACHE_KEY_PREFIX': 'app_explorer',
-                            'CACHE_OPTIONS': {'COMPRESS_MIN_LENGTH': 5000, 'COMPRESS_LEVEL': 6}
-                            })
+DB_CACHE_CONFIGS = os.getenv(
+    'DB_CACHE_CONFIGS',
+    # {
```
jonespm, line 159: Should remove this commented code

backend/canvas_app_explorer/lti1p3.py
jonespm: Nevermind CacheConfig is a class-like object so it's fine named like this.

backend/settings.py
```diff
@@ -154,12 +153,19 @@
...
-                            'LOCATION': 'canvas_app_explorer_cache',
-                            'CACHE_KEY_PREFIX': 'app_explorer',
-                            'CACHE_OPTIONS': {'COMPRESS_MIN_LENGTH': 5000, 'COMPRESS_LEVEL': 6}
-                            })
+DB_CACHE_CONFIGS = os.getenv(
+    'DB_CACHE_CONFIGS',
+    {
+        'CACHE_TTL': 600,
+        'BACKEND': 'django_redis.cache.RedisCache',
+        'LOCATION': f"redis://{os.getenv('REDIS_HOST', 'redis')}:6379/1",
```
jonespm, line 161: I believe this should be more configurable and documented. 

The port and database number probably should be just part of the string. 
```suggestion
        'LOCATION': os.getenv('REDIS_URL', 'redis://redis:6379/1'),
```

.env.sample
```diff
@@ -61,6 +61,10 @@ ALLOWED_HOSTS=.ngrok.io,.localhost,127.0.0.1
 # Database Password
 # DB_PASSWORD=cae_pw
 
+##### Redis
+
+REDIS_PASS=
```
jonespm, line 66:
```suggestion
# Redis Password, default is blank
REDIS_PASS=
# Redis URL, full url of redis server and database number, has a default in the code
# REDIS_URL= redis://redis:6379/1
```

backend/settings.py
jonespm: I updated the documentation to reflect this.

PR #278, opened 2022-10-06
added 'blank=True' attribute to internal_notes, so that a LTI tool ca…

roujiawang:
…rd can be created with a blank internal note field


backend/canvas_app_explorer/migrations/0011_ltitool_internal_notes.py
```diff
@@ -14,6 +14,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='ltitool',
             name='internal_notes',
-            field=tinymce.models.HTMLField(help_text='a place to put helpful info for admins, not visible to users', null=True),
+            field=tinymce.models.HTMLField(blank=True, help_text='a place to put helpful info for admins, not visible to users', null=True),
```
ssciolla, line 17: Hi @roujiawang, because your first change was already committed and moved to our deployments, we think it's better to generate a new migration -- rather than modify an existing one. That way we don't have to intervene and rewind history in our databases to get it to work (i.e. we've already applied the earlier version).

ssciolla (changes requested):

backend/canvas_app_explorer/migrations/0011_ltitool_internal_notes.py
roujiawang: I have reverted the previous commit and made a new commit that includes a new migration file.

ssciolla (approved):

PR #285, opened 2022-10-18
Convert likely string course ID to an integer (#284)

### LINKED ISSUE #284: Course ID LTI variable is now a string
Following the latest Canvas [changes](https://community.canvaslms.com/t5/Canvas-Change-Log/Canvas-Platform-Breaking-Changes/ta-p/262015), LTI variable substitution values are always going to be strings. Unfortunately, the current version of this tool expects an integer and throws an error if it isn't one.

I guess ideally, a fix to this problem should handle both possible types, converting to `int` as necessary.

ssciolla:
This PR aims to resolve #284.

### DIFF
--- a/backend/canvas_app_explorer/lti1p3.py
+++ b/backend/canvas_app_explorer/lti1p3.py
@@ -26,6 +26,15 @@ DUMMY_CACHE = 'DummyCache'
 
 TOOL_CONF = DjangoDbToolConf()
 
+# Custom LTI variable keys
+USERNAME_KEY = 'user_username'
+COURSE_ID_KEY = 'canvas_course_id'
+COURSE_ROLES_KEY = 'canvas_course_roles'
+
+
+class LTILaunchError(Exception):
+    pass
+
 
 # do not require deployment ids if LTI_CONFIG_DISABLE_DEPLOYMENT_ID_VALIDATION is true
 class ExtendedDjangoMessageLaunch(DjangoMessageLaunch):
@@ -114,14 +123,28 @@ def create_user_in_django(request: HttpRequest, message_launch: ExtendedDjangoMe
     custom_params = launch_data['https://purl.imsglobal.org/spec/lti/claim/custom']
     logger.debug(f'lti_custom_param {custom_params}')
     if not custom_params:
-        raise Exception(
-            f'You need to have custom parameters configured on your LTI Launch. Please see the LTI installation guide on the Github Wiki for more information.'
+        raise LTILaunchError(
+            'You need to have custom parameters configured on your LTI Launch. ' +
+            'Please see the LTI installation guide on the Github Wiki for more information.'
         )
+
+    custom_param_keys = custom_params.keys()
+    if not (
+        USERNAME_KEY in custom_param_keys and
+        COURSE_ID_KEY in custom_param_keys and
+        COURSE_ROLES_KEY in custom_param_keys
+    ):
+        raise LTILaunchError(
+            'One or more required custom LTI variables were not defined. ' +
+            f'These variables are required: {", ".join([USERNAME_KEY, COURSE_ID_KEY, COURSE_ROLES_KEY])}'
+        )
+
     course_name = launch_data['https://purl.imsglobal.org/spec/lti/claim/context']['title']
     roles = launch_data['https://purl.imsglobal.org/spec/lti/claim/roles']
-    username = custom_params['user_username']
-    course_id = custom_params['canvas_course_id']
-    course_roles = custom_params['canvas_course_roles'].split(',')
+
+    username = custom_params[USERNAME_KEY]
+    course_id = custom_params[COURSE_ID_KEY]
+    course_roles = custom_params[COURSE_ROLES_KEY].split(',')
 
     if 'email' not in launch_data.keys():
         logger.warn('An instructor/admin likely launched the tool using Student View (Test Student).')
@@ -162,11 +185,11 @@ def create_user_in_django(request: HttpRequest, message_launch: ExtendedDjangoMe
         try:
             course_id_int = int(course_id)
         except ValueError:
-            raise Exception(f'Course ID from LTI launch cannot be converted to an integer. Value: {course_id}')
+            raise LTILaunchError(f'Course ID from LTI launch cannot be converted to an integer. Value: {course_id}')
 
         request.session['course_id'] = course_id_int
     else:
-        raise Exception(f'Course ID from LTI launch cannot be null.')
+        raise LTILaunchError(f'Course ID from LTI launch cannot be null.')
 
 
 @csrf_exempt
--- a/backend/canvas_app_explorer/lti1p3.py
+++ b/backend/canvas_app_explorer/lti1p3.py
@@ -158,17 +158,15 @@ def create_user_in_django(request: HttpRequest, message_launch: ExtendedDjangoMe
     user_obj.backend = 'django.contrib.auth.backends.ModelBackend'
     django_login(request, user_obj)
 
-    course_id_error = Exception('The canvas_course_id custom LTI variable must be configured.')
     if course_id is not None:
         try:
             course_id_int = int(course_id)
         except ValueError:
-            logger.error(f'Course ID from LTI launch cannot be converted to an integer: {course_id}')
-            raise course_id_error
+            raise Exception(f'Course ID from LTI launch cannot be converted to an integer. Value: {course_id}')
 
         request.session['course_id'] = course_id_int
     else:
-        raise course_id_error
+        raise Exception(f'Course ID from LTI launch cannot be null.')
 
 
 @csrf_exempt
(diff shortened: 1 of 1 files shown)


backend/canvas_app_explorer/lti1p3.py
```diff
@@ -158,10 +158,17 @@ def create_user_in_django(request: HttpRequest, message_launch: ExtendedDjangoMe
     user_obj.backend = 'django.contrib.auth.backends.ModelBackend'
     django_login(request, user_obj)
 
-    if course_id is not None and isinstance(course_id, int):
-        request.session['course_id'] = course_id
+    course_id_error = Exception('The canvas_course_id custom LTI variable must be configured.')
+    if course_id is not None:
```
ssciolla, line 162: Not exactly sure this would ever come up in practice, it'd be more likely than we'd get `KeyError` with `course_id` being undefined. However, it is technically possible, since it is `None` if you set the value of the `course_id` key to `null` in the LTI config.

backend/canvas_app_explorer/lti1p3.py
```diff
@@ -158,10 +158,17 @@ def create_user_in_django(request: HttpRequest, message_launch: ExtendedDjangoMe
...
+        try:
+            course_id_int = int(course_id)
+        except ValueError:
+            logger.error(f'Course ID from LTI launch cannot be converted to an integer: {course_id}')
+            raise course_id_error
```
zqian, line 167: I feel like the error msg should be different than the course_id_error. We can use the same message put in the log: 'Course ID from LTI launch cannot be converted to an integer: {course_id}'

backend/canvas_app_explorer/lti1p3.py
```diff
@@ -158,10 +158,17 @@ def create_user_in_django(request: HttpRequest, message_launch: ExtendedDjangoMe
     user_obj.backend = 'django.contrib.auth.backends.ModelBackend'
     django_login(request, user_obj)
 
-    if course_id is not None and isinstance(course_id, int):
-        request.session['course_id'] = course_id
+    course_id_error = Exception('The canvas_course_id custom LTI variable must be configured.')
```
jonespm, line 161: Maybe a nitpick, but feels like we should define a custom exception class or maybe use an `LtiException` here. https://github.com/dmitry-viskov/pylti1.3/blob/master/pylti1p3/exception.py

And these aren't exactly the same error.

In the first raise it's a value error that course_id is not defined as an integer
The second is that it's not configured at all. 

So seems like 2 separate exceptions.

jonespm: I felt like this was a little of the problem with the existing code, one message was catching for two different errors, but they were different errors and it was a little hard to tell which was failing.

backend/canvas_app_explorer/lti1p3.py
ssciolla: Okay, see https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/285/commits/da76ae33b3a6e57a95cf4a4addc15f53689854ce.

backend/canvas_app_explorer/lti1p3.py
ssciolla: Okay, this partially addresses this: https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/285/commits/da76ae33b3a6e57a95cf4a4addc15f53689854ce

I'm somewhat opposed to using `pylti1p3.LtiException` since the exception is not originating from that library. I could create a custom exception class if you feel strongly about this. That seems like it would be more valuable if we planned to catch and do something about these errors.

Just a point of clarification: the `is not None` is not actually catching if the variable is not set up at all. That would just be a `KeyError`, I believe. We aren't doing this for any of the other custom variables, but I could also add that if we want to strengthen this a bit now.

jonespm: I'm fine leaving it. I know we created CanvasHTTPError and maybe the errors in here should have been something. 

They don't originate from the library but they are related to errors with the library. 

The Django-LTI is using [LtiException](https://github.com/academic-innovation/django-lti/blob/main/lti_tool/middleware.py).

And I feel like we've raised exceptions from other packages like GraphQL, Django and maybe Pandas. Not sure if there's a correct answer here. I guess it just makes it easier to catch, if necessary, if we have specific exceptions. But maybe these are just uncaught.

jonespm (approved):

backend/canvas_app_explorer/lti1p3.py
ssciolla: @jonespm, does this look better? I can revert if not. https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/285/commits/372fb0f679206c1cc7f1b953ad2ab28fcbd31f52

zqian (approved):

backend/canvas_app_explorer/lti1p3.py
```diff
@@ -114,14 +123,28 @@ def create_user_in_django(request: HttpRequest, message_launch: ExtendedDjangoMe
...
-        raise Exception(
-            f'You need to have custom parameters configured on your LTI Launch. Please see the LTI installation guide on the Github Wiki for more information.'
+        raise LTILaunchError(
+            'You need to have custom parameters configured on your LTI Launch. ' +
+            'Please see the LTI installation guide on the Github Wiki for more information.'
+        )
+
+    custom_param_keys = custom_params.keys()
+    if not (
+        USERNAME_KEY in custom_param_keys and
```
jonespm, line 133: I believe this [could be written](https://www.geeksforgeeks.org/python-check-if-given-multiple-keys-exist-in-a-dictionary/) as

`if not (set([USERNAME_KEY, COURSE_ID_KEY, COURSE_ROLES_KEY]).issubset(custom_params.keys())):`

You could probably also just try/except a KeyError later on 145. I prefer to write EAFP in Python rather than verifying everything. But that's an individual preference. Just trying to make have this a little less verbose.

backend/canvas_app_explorer/lti1p3.py
jonespm: Some of the verification changes could probably be changed to be more programatic but up to you. I'm fine with with this and the exception looks good.

But a small nitpick, I was reading that some books like [Python in a Nutshell](https://www.oreilly.com/library/view/python-in-a/0596100469/ch06s05.html) and [Stack Overflow ](https://stackoverflow.com/a/26938914/3708872) recommend using a docstring rather than a pass in custom exceptions that do nothing. They call it "maybe perfect". :)

jonespm (approved):

backend/canvas_app_explorer/lti1p3.py
ssciolla: Yeah, I like your version better. I actually looked at `set` before cause it felt like I could do something there, but I missed this method.

I need to do some more research and soul-searching on EAFP. I think my instinct is to look before you leap (LBYL) because I find it more readable and have been burned by imprecise `try`/`except`s before, but there's probably a time and a place for both.

backend/canvas_app_explorer/lti1p3.py
ssciolla: https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/285/commits/5f21b063c6066d60d5459afe5e495b6a9ab80005

backend/canvas_app_explorer/lti1p3.py
ssciolla: https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/285/commits/5f21b063c6066d60d5459afe5e495b6a9ab80005

ssciolla: I read/skimmed this article just now: https://realpython.com/python-lbyl-vs-eafp/. It has some good points about using EAFP in Python, especially to minimize extra checks and prevent race condition issues.

jonespm: In Java, you basically have to NPE check almost everything but otherwise it's usually just exception handling everywhere. 

Interesting article and the point about the [zero-cost exceptions](https://github.com/python/cpython/issues/84403) is something I didn't know Python didn't have. So basically Java and Python have very inexpensive exception handling. 

I just see patterns like the one here and they seem repetitive and redundant, where they could just be solved with a try/except. This is the case for readability and repetition in the article. And there are cases to use both and not be all-in.

I think this case would fall into the uncommon and unlikely to fail side of the column. 

Use LBYL for | Use EAFP for
-- | --
Operations that are likely to fail | Operations that are unlikely to fail
Irrevocable operations, and operations that may have a side effect | Input and output (IO) operations, mainly hard drive and network operations
Common exceptional conditions that can be quickly prevented beforehand | Database operations that can be rolled back quickly

ssciolla: Definitely room for me to grow here. I'll have to be more mindful of my options for exceptions in the future.

PR #287, opened 2022-11-29
wrote a new command to fetch the information of available external to…

roujiawang:
…ols from Canvas

### DIFF
--- a/backend/canvas_app_explorer/management/commands/pull_canvas_tools.py
+++ b/backend/canvas_app_explorer/management/commands/pull_canvas_tools.py
@@ -1,0 +1,41 @@
+from django.core.management.base import BaseCommand, CommandParser
+from canvasapi import Canvas
+
+from backend.canvas_app_explorer.models import LtiTool, CanvasPlacement
+
+# register customized actions with manage.py
+# usage: python manage.py pull (<num_tools_retrived>?) <API_KEY> (<API_URL>) 
+class Command(BaseCommand):
+    help = "pull info about available external tools on Canvas to populate a local database"
+
+    def add_arguments(self, parser):
+        parser.add_argument('api_key', nargs=1, type=str)
+        parser.add_argument('api_url', nargs='?', type=str, default="https://umich-dev.instructure.com")
+
+    def handle(self, *args, **options):
+        
+        API_KEY = options['api_key'][0]
+        API_URL = options['api_url']
+        canvas = Canvas(API_URL, API_KEY)
+
+        primary_acount = canvas.get_account(1)
+        ex_tool_list = primary_acount.get_external_tools()
+        # for ManyToManyField setup
+        course_nav_placement = CanvasPlacement.objects.get(name='Course Navigation')
+        for ex_tool in ex_tool_list:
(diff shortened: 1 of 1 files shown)


backend/canvas_app_explorer/management/commands/pull_canvas_tools.py
```diff
@@ -0,0 +1,41 @@
...
+        parser.add_argument('api_key', nargs=1, type=str)
+        parser.add_argument('api_url', nargs='?', type=str, default="https://umich-dev.instructure.com")
+
+    def handle(self, *args, **options):
+        
+        API_KEY = options['api_key'][0]
+        API_URL = options['api_url']
+        canvas = Canvas(API_URL, API_KEY)
+
+        primary_acount = canvas.get_account(1)
```
jonespm, line 21: Should probably make this account configurable. It want to be run on subaccounts.

backend/canvas_app_explorer/management/commands/pull_canvas_tools.py
```diff
@@ -0,0 +1,41 @@
...
+        course_nav_placement = CanvasPlacement.objects.get(name='Course Navigation')
+        for ex_tool in ex_tool_list:
+            # for now, only care about tools with placement "course_navigation"
+            if ex_tool.__dict__["course_navigation"] is not None \
+                and ex_tool.__dict__["course_navigation"]["enabled"] == True:
+
+                # avoid duplicates: get_or_create
+                if not LtiTool.objects.filter(canvas_id=ex_tool.id).exists():
+                    tool = LtiTool.objects.create(
+                        name=ex_tool.name,
```
jonespm, line 33: I think we'll want to append something to the name like "(PLACEHOLDER)".

We're also using the fact that the ID not being set means the tool is "disabled" now, we probably will need to add a new column to indicate this if we're having a new process to run this.

backend/canvas_app_explorer/management/commands/pull_canvas_tools.py
```diff
@@ -0,0 +1,41 @@
...
+        API_URL = options['api_url']
+        canvas = Canvas(API_URL, API_KEY)
+
+        primary_acount = canvas.get_account(1)
+        ex_tool_list = primary_acount.get_external_tools()
+        # for ManyToManyField setup
+        course_nav_placement = CanvasPlacement.objects.get(name='Course Navigation')
+        for ex_tool in ex_tool_list:
+            # for now, only care about tools with placement "course_navigation"
+            if ex_tool.__dict__["course_navigation"] is not None \
```
jonespm, line 27: We'll probably also want to make this configurable but this is fine hardcoding for now.

jonespm:
Good start on this feature idea! I'm sure we'll have a few more things to talk about with this and to get through the other ToDo items.

backend/canvas_app_explorer/management/commands/pull_canvas_tools.py
ssciolla: @roujiawang added an `is_public` field now, so I think we're good on the second front. It might be useful to include placeholder text, not completely sold though.

Issue #302, opened 2024-11-06
docker build problem with netcat package


zqian:
The `docker compose build` generates the following error with `netcat` package installation:

```
#0 3.391 Package netcat is a virtual package provided by:
#0 3.391   netcat-openbsd 1.219-1
#0 3.391   netcat-traditional 1.10-47
#0 3.391
#0 3.395 E: Package 'netcat' has no installation candidate
------
failed to solve: executor failed running [/bin/sh -c apt-get update &&     apt-get install -y --no-install-recommends     build-essential default-libmysqlclient-dev netcat vim-tiny jq python3-dev git supervisor curl &&     apt-get upgrade -y &&     apt-get clean -y &&     rm -rf /var/lib/apt/lists/*]: exit code: 100
```

PR #303, opened 2024-11-06
#302 updated the netcat package name

zqian:


jonespm (approved):

dockerfiles/Dockerfile
```diff
@@ -7,7 +7,15 @@ FROM python:3.8-slim
 COPY requirements.txt .
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-    build-essential default-libmysqlclient-dev netcat vim-tiny jq python3-dev git supervisor curl && \
+    build-essential \
+    default-libmysqlclient-dev \
+    netcat-openbsd \
+    vim-tiny \
```
jonespm, line 13: We probably could remove vim-tiny but maybe that's a separate issue

zqian: Let's leave that as a separate issue.

PR #306, opened 2024-11-06
Create dependabot.yml to test version updates

jonespm:


.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
...
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  # Maintain dependencies for python
+  - package-ecosystem: "pip" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "monthly"
```
zqian, line 12: I think the interval could be increased to weekly

.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
...
+  # Maintain dependencies for npm
+  - package-ecosystem: "npm"
+    directory: "/frontend"
+    schedule:
+      interval: "monthly"
```
zqian, line 17: same here, let's increase the interval to weekly

zqian (approved):
Approved the autogenerated PR, with a comment for increasing the interval to "weekly"

.github/dependabot.yml
jonespm: Okay, I can change and try it weekly to see how it works. 

But if we aren't doing very frequent releases and automated testing I don't think it will matter and will just be a lot of updates that we aren't deploying. We can see how it goes.

PR #319, opened 2024-11-06
Configuring to ignore non LTS versions

jonespm:
5.2 LTS isn't released until later in 2025 so going to see if it keeps it back to 4.2 releases

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,7 +15,7 @@ updates:
     ignore:
       - dependency-name: "django"
         versions:
-          - ">= 4.3"  # This ignores versions above 4.2, allowing only 4.2.x LTS upgrades
+          - "> 4.2"  # This ignores versions above 4.2, allowing only 4.2.x LTS upgrades
   # Maintain dependencies for npm
   - package-ecosystem: "npm"
     directory: "/frontend"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,12 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+    allow:
+      - dependency-type: "direct"
+    ignore:
+      - dependency-name: "django"
+        versions:
+          - ">= 4.3"  # This ignores versions above 4.2, allowing only 4.2.x LTS upgrades
   # Maintain dependencies for npm
   - package-ecosystem: "npm"
     directory: "/frontend"


zqian:
When reviewing the code, I found the following changes are also needed in the dockerfiles/Dockerfile:

```
diff --git a/dockerfiles/Dockerfile b/dockerfiles/Dockerfile
index 48c19b2..e1c3aa7 100644
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -1,7 +1,7 @@
 # FROM directive instructing base image to build upon
 # This could be used as a base instead:
 # https://hub.docker.com/r/nikolaik/python-nodejs
-FROM python:3.8-slim
+FROM python:3.10-slim-bullseye

 # NOTE: requirements.txt not likely to change between dev builds
 COPY requirements.txt .
diff --git a/requirements.txt b/requirements.txt
index 786d0fc..be4a0a8 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ gunicorn==20.1.0
 mysqlclient==2.1.1

 # Django and related
-Django==3.2.16
+Django==4.2.16
 django-csp==3.7 # For Content Security Policy
 django-db-file-storage==0.5.5 # Support for storage in the database
 django-mysql==4.15.0
(END)
```

jonespm:
I think @pushyamig said she was going to put in a PR to upgrade the python version. 

Master has Django updated, this branch might not be in sync.

pushyamig:
I have upgraded python 3.10 locally. It built fine, and starts fine but cannot launch the app since the read me did not had info how to add LTI config to
The tool once I have that figured out thae I will open a PR

zqian:
The README contains this line: "This app can currently only be launched via LTI. Please see the Wiki for instructions on configuring with LTI." So you can find the LTI installation instruction inside the wiki page: https://github.com/tl-its-umich-edu/canvas-app-explorer/wiki/Configuring-LTI-1.3

zqian:
@jonespm would you please sync the repo now that Pushyami has merged in https://github.com/tl-its-umich-edu/canvas-app-explorer/pull/321?

.github/dependabot.yml
```diff
@@ -10,6 +10,12 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
+    allow:
+      - dependency-type: "direct"
+    ignore:
+      - dependency-name: "django"
```
jonespm, line 16: In this project's requirements it has Django. pip is supposed to be case insensitive so not 100% sure if this will work or not. Can just update it now to be capitalized. 
```suggestion
      - dependency-name: "Django"
```

PR #321, opened 2024-11-08
#320 upgrading to python 3.10 and Canvas Oauth pkg

### LINKED ISSUE #320: Update Python version to 3.10

pushyamig:
Fixes #320 

I am using the same Python 3.10 version that Myla is using.

I was able to build, start and launch the tool

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -1,7 +1,7 @@
 # FROM directive instructing base image to build upon
 # This could be used as a base instead: 
 # https://hub.docker.com/r/nikolaik/python-nodejs
-FROM python:3.8-slim
+FROM python:3.10-slim-bookworm
 
 # NOTE: requirements.txt not likely to change between dev builds
 COPY requirements.txt .
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,4 +26,4 @@ PyLTI1p3==1.12.1
 canvasapi==3.3.0
 
 # Not in pypi https://github.com/Harvard-University-iCommons/django-canvas-oauth
-https://github.com/Harvard-University-iCommons/django-canvas-oauth/archive/v1.1.0.tar.gz
+https://github.com/Harvard-University-iCommons/django-canvas-oauth/archive/v1.1.1.tar.gz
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -26,4 +26,4 @@ PyLTI1p3==1.12.1
 canvasapi==3.3.0
 
 # Not in pypi https://github.com/Harvard-University-iCommons/django-canvas-oauth
-https://github.com/Harvard-University-iCommons/django-canvas-oauth/archive/v1.1.0.tar.gz
+https://github.com/Harvard-University-iCommons/django-canvas-oauth/archive/v1.1.1.tar.gz
```
pushyamig, line 29: This package support reference added only  Python 3.6 and Django > 2.0.  https://github.com/Harvard-University-iCommons/django-canvas-oauth/blob/master/setup.py

I guess the support range is large, the pkg got a recent update in 2024. When I did not update to latest I saw this error "TypeError: get_random_string() missing 1 required positional argument: 'length'' 

Looking at the latest commit from this repo they fixed exactly this https://github.com/Harvard-University-iCommons/django-canvas-oauth/commit/299997426b3f9ca830e7d3085f70aba6fc92504c

SO I guess as long as it build/runs it should be ok with the Python/Django

dockerfiles/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # FROM directive instructing base image to build upon
 # This could be used as a base instead: 
 # https://hub.docker.com/r/nikolaik/python-nodejs
-FROM python:3.8-slim
+FROM python:3.10-slim-bookworm
```
zqian, line 4: This is version used by the MyLA. ROHQ uses `python:3.10-slim-bullseye`. Maybe we can adopt the same model among all T&L apps

zqian (approved):
docker compose builds successfully. Left a comment on the Python package.

dockerfiles/Dockerfile
jonespm: Agreed, bullseye [is no longer supported](https://endoflife.date/debian). MyLA was [upgraded to bookworm](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1513) last year, so an issue should be created to upgrade ROHQ and any place else this is noticed.

PR #331, opened 2024-11-08
I322 gh actions

### LINKED ISSUE #322: Add Github Action workflow for Docker Build
This step is pretty much same as other TL repo Projects.

pushyamig:
Fixes #322 

I have built this on my forked repo. 
https://github.com/pushyamig/canvas-app-explorer/actions/runs/12122482979/job/33795762679

Now with this change, we opted for only one docker file and supervisor will be using in prod as well. Only Backend will be started in Prod and not Frontend


.github/workflows/cae.yml
```diff
@@ -0,0 +1,65 @@
...
+        description: 'Version'
+        required: true
+        default: 'YYYY.MINOR.MICRO'
+env:
+  REPO_URL: ${{ github.repository }}
+    
+jobs:
+  build:
+    # to test a feature, change the repo name to your github id
+    if: github.repository_owner == 'tl-its-umich-edu' || github.event_name == 'workflow_dispatch'
```
pushyamig, line 24: I am pointing to TL repo only since I did not want to forked repo for a build ( this step is common to all TL Repos)

`workflow_dispatch` will let us do a manual trigger of the build

dockerfiles/Dockerfile.prod
```diff
@@ -1,6 +1,6 @@
 # node-build stage
 
-FROM docker-registry.default.svc:5000/openshift/node:16-slim AS node-build
+FROM node:20-slim AS node-build
```
pushyamig, line 3: @jonespm I chose Node 20 since it is under LTS https://endoflife.date/nodejs.

CCM also is on Node 20 until rewrite done

pushyamig: Also, @jonespm I have a question, in the Local Supervisor setup (supervisor_docker.conf) or package.json, there is no mention of Node version. So I am guessing it is picking up what ever based on the default set with supervisor?

Should Be set a Node version in supervisor_docker.conf? since that's how the FE package are build from

dockerfiles/Dockerfile.prod
```diff
@@ -1,6 +1,6 @@
 # node-build stage
```
pushyamig, line 1: I removed all the openshift references, since we will be switching to doing builds on Github and not from Openshift any more.

So this mean we have some workflow change to existing process on Openshift Deployment. I am guessing should be minor and the process has been established from other TL project in Openshift

.github/workflows/cae.yml
```diff
@@ -0,0 +1,65 @@
+name: CAE Build/Release
```
zqian, line 1: since we are rebranding the CAE tool to IPT, can we rename this file to ipt.yml? and replace all occurrence of `cae` in this PR?

zqian (approved):

.github/workflows/cae.yml
pushyamig: I can do that if that's the direction we are moving toward with naming this app and project

dockerfiles/Dockerfile.prod
jonespm: Correct, it just runs whatever npm version is installed by this process. I don't think it needs a node version in there.

dockerfiles/Dockerfile.prod
jonespm: I think with this process the Dockerfile.openshift/Dockerfile.prod can just be deleted. That was only if this needed to build on Openshift and we'll just switch that to pull from GH

pushyamig: @jonespm , I don't think we can delete `openshift/Dockerfile.prod` since 

Dockerfile: CMD ["/usr/bin/supervisord", "-c", "/code/deploy/supervisor_docker.conf"]
Dockerfile.Prod: CMD ["/code/start_backend.sh"]

`RUN npm run build:frontend`,  This command is missing from `Dockerfile` I don't see JS Bundle is made with Dockerfile . Here is the output from that build https://github.com/pushyamig/canvas-app-explorer/actions/runs/11746094247/job/32725028305

.github/workflows/cae.yml
pushyamig: I rename CAE reference to IPT

dockerfiles/Dockerfile.prod
pushyamig: @jonespm Let me know if you have any further concerns on this change. I will work to resolve it.

jonespm: I see what you mean. It would be fine running it through supervisord in prod too (possibly better) but we don't need to run both the frontend and backend processes since the frontend is static. 

So we'd need a to set it up so it only starts the backend but starts the frontend on development. 

I think the way this would be configured would be to just have the backend configured with `autostart=true` and the frontend be `autostart=false`. Then either through docker-compose or the startup script also start the frontend.

.github/workflows/ipt.yml
jonespm, line 1: Maybe we should make this a more generic like build_release.yml so we can avoid possible name changes? And update the name on the first line.

.github/workflows/ipt.yml
```diff
@@ -0,0 +1,65 @@
+name: IPT Build/Release
+
+on:
+  push:
+    # takes muliple branch names
+    branches: 
+      - main
+      - '[0-9][0-9][0-9][0-9].[0-9][0-9].*' # 2021.01.x
+    tags:
+      - '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9][0-9]' # 2021.01.01
```
jonespm, line 10: We actually don't need this to build when pushing to a numbered tag since we'll be triggering that manually.

.github/workflows/ipt.yml
```diff
@@ -0,0 +1,65 @@
...
+            username: ${{ github.actor }}
+            password: ${{ secrets.GITHUB_TOKEN }}
+      
+      - name: Push Docker image to GitHub Container Registry
+        run: |
+          docker push ghcr.io/${{ env.REPO_URL }}:${BRANCH_NAME}
+    
+  release: 
+     # Making sure that release only runs for tag pushes
+    if: startsWith(github.ref, 'refs/tags/') && github.repository_owner == 'tl-its-umich-edu'
```
jonespm, line 52: This check for repository_owner doesn't need to be in this step.
```suggestion
    if: startsWith(github.ref, 'refs/tags/')'
```

jonespm:
We may want to wait on merging this for at least a week or so as I have a few more thoughts on this. It's not _required_ to test or even release to prod using the current workflow and doing it will involve some more changes to Openshift. 

* I'd probably delete the Dockerfile.prod and make the Dockerfile run supervisor.
* Supervisor should only start the backend by default
* The frontend should be started either by docker-compose.yml (which is used locally) or by start_backend.sh if it running locally. Some variable would have to be passed, possibly from docker-compose.yml to indicate that it's running locally
* Then we'd need to make changes to the Openshift to pull Imagestreams from this new image.

dockerfiles/Dockerfile.prod
pushyamig: So add environmental variable  in docker compose and allow it to run FE and BE during dev not in the prod

pushyamig:
>We may want to wait on merging this for at least a week or so as I have a few more thoughts on this. 

Ok, I will think about your ideas and how to implement it. The Supervisor stuff is new to me so it will be good learning

jonespm:
Sure! It looks like the website might be down. It basically allows you to start up multiple processes in a single container/pod rather than running multiple. 

When developing this app, I thought this would be more efficient for running in prod due to cost, but isn't a huge advantage locally. And even in prod there's only one process running. Though other advantages are it can attempt to restart the app quicker if it fails and you have better configuration of where the logs are going.

pushyamig:
This PR is functional but need to test the Build created from GH action on Openshif, since there is a slight change in how the app starts since now supervisor is used to launch.

I will keep the PR in Draft mode, once we pass the Prod release phase I will have it tested and open the PR

pushyamig:
I don't think this has anything further to do. We just need to make sure if OpenShift pod starts up fine with Supervior. That's the only thing I couldn't test. If you are ok with testing that later, we can merge it.

jonespm:
Okay, cool, I think we'll just have to test it there. I'd like to have everything in Github packages for the new ImageStream project idea and just noticed that this was still out there and it didn't have one. I believe we'll probably have a new release of this tool near the end of the semester.

jonespm (approved):

PR #334, opened 2024-11-08
#333 supporting local debugging

### LINKED ISSUE #333: Support Local debugging with debugpy
I felt it was somewhat essential for debugging this project

pushyamig:
Fixes #333 

Felt having this make it little bit easier for local debugging for people who join new to the project

### DIFF
--- a/backend/debugpy.py
+++ b/backend/debugpy.py
@@ -1,0 +1,15 @@
+import os
+import debugpy
+import logging
+
+def config_to_bool(value): 
+    return str(value).lower() in ('true', '1', 'yes', 'on')
+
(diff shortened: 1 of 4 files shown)


jonespm (approved):
This seems fine, it looks like there were some thoughts to enable this but nothing was ever done.

jonespm (changes requested):
Can you wait a minute on this? I think we had a simpler pattern that Zhen found for debugging in ROHQ

jonespm (approved):
Nevermind, I'm not sure if there's a better way to do it with gunicorn, only with uvicorn. 

I think this looks good.

PR #347, opened 2024-11-12
Fixes #346 - Continued dependabot to group scheduled updates

### LINKED ISSUE #346: Configured dependabot to group scheduled updates
It would be easier to handled scheduled updates if they were grouped.

Here's the example of the configuration that needs to be made.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#grouping-dependabot-updates-into-one-pull-request

jonespm:
This seems like it should be enough to configure the grouped dependencies. This groups everything. This could be further configured to create more specific groups.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,8 +16,14 @@ updates:
       - dependency-name: "django"
         versions:
           - "> 4.2"  # This ignores versions above 4.2, allowing only 4.2.x LTS upgrades
+    groups:
+      all-python-dependencies:
+        pattern: "*"
   # Maintain dependencies for npm
   - package-ecosystem: "npm"
     directory: "/frontend"
     schedule:
       interval: "weekly"
+    groups:
+      all-npm-dependencies:
+        pattern: "*"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -16,8 +16,14 @@ updates:
       - dependency-name: "django"
         versions:
           - "> 4.2"  # This ignores versions above 4.2, allowing only 4.2.x LTS upgrades
+    groups:
+      all-python-dependencies:
+        pattern: "*"
```
zqian, line 21: In the[ referenced resource page](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#grouping-dependabot-updates-into-one-pull-request), only `patterns` is used, not pattern. I am not sure whether we need to change to:

`patterns:
-*
`

zqian (approved):

jonespm:
Good catch, I didn't test this so will change it to the plural version.

.github/dependabot.yml
jonespm: Yes, this is changed thanks!

PR #348, opened 2024-11-13
Dependency Update for the dependabot PRs

pushyamig:
For all the PR #345 #344 #343 #342 #342 #341 #340 #339 #328 #327 opened by dependabot this is combined PR since some of the upgrades caused build and run issues. 

Once this PR is approve I will close those issues


pushyamig:
The Goal of Dependency update is to keep semi-update and not get the latest but LTS packages. We are picking and choosing based on the library and it is random.

I am going to get a separate PR for upgrading React to 18 and probably Typescript to 5. My anticipation it might a good trade-off at this point. I am going to hold back on upgrading some libraries like  MUI to v6

pushyamig:
This node package cli can `ncu` gives a quick update what is the latest and where we are at 

https://www.npmjs.com/package/npm-check-updates

zqian (approved):
Testes successfully on Canvas Test for both UI functions and also swagger-ui endpoints, authorized with session_id

PR #350, opened 2024-11-13
issue_335 changes to tool labels and button positions

zqian:


pushyamig:
I have updated this branch from latest from main

frontend/app/components/ToolCard.tsx
```diff
@@ -119,6 +119,14 @@ export default function ToolCard (props: ToolCardProps) {
...
           aria-busy={updateToolNavLoading}
         >
+          <Button
+            onClick={() => setShowMoreInfo(!showMoreInfo)}
+            aria-expanded={showMoreInfo}
+            aria-label={`Show ${moreOrLessText} Info`}
+            startIcon={!showMoreInfo ? <ExpandMoreIcon /> : <ExpandLessIcon />}
+          >
+            {moreOrLessText}
+          </Button>
```
pushyamig, line 129: Minor detail: The Mock up now say `Learn more`  change reflect `More`. don't know if we are sticking with existing design from CAE as discussed in the Slack yesterday. Keeping the changes to minimal. I am ok if that's the case

pushyamig (approved):
Looks Good

frontend/app/components/ToolCard.tsx
zqian: I will keep the "More" vs "Less" labels for now. Thanks!

jonespm: I'd have to look at the notes but I know we had a lot of UX back and forth about this button. I also believe it had an accessibility issue and I'm not sure if it was resolved. I'll have to see it when it's running, but that would be a separate issue.

jonespm (approved):

PR #355, opened 2024-11-14
Upgrade node and Dockerfile to get to build on Openshift

### LINKED ISSUE #354: Upgrade node and Dockerfiles to get to build on Openshift
The current `Dockerfile.openshift` does not build on Openshift because the images are outdated and not present and the registry is pointing to the old location.

Even though this might be removed in #322 or other issues I think we should do a quick fix so we can get it building. I tested locally and Node 20 seems to work fine.

jonespm:
Fixes #354
The 3.10-slim is the same as 3.10-slim-bookworm, but this one exists in Openshift

### DIFF
--- a/dockerfiles/Dockerfile.openshift
+++ b/dockerfiles/Dockerfile.openshift
@@ -14,10 +14,15 @@ FROM image-registry.openshift-image-registry.svc:5000/openshift/python:3.10-slim
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-    build-essential default-libmysqlclient-dev netcat vim-tiny jq python3-dev git supervisor curl && \
+    build-essential netcat-openbsd vim-tiny jq python3-dev git supervisor curl pkg-config && \
     apt-get upgrade -y && \
     apt-get clean -y && \
     rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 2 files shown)


jonespm:
@pushyamig This will cause conflicts to #331 but this will also get what we have building and running on Openshift for now.

dockerfiles/Dockerfile.openshift
```diff
@@ -1,6 +1,6 @@
 # node-build stage
```
pushyamig, line 2: These changes needs to be implemented in the file since I did not made this .openshift as I felt it would be removed. 

https://github.com/tl-its-umich-edu/instructor-productivity-tools/pull/348/files

pushyamig: After the mysql version upgrade the docker build was failing so I needed to update it as shown in the PR.

pushyamig (approved):
I suggested the changes for change on Dockerfile for openshift. other than that this PR looks OK. Some point we might be deleting Openshift DOcker file and that is file

dockerfiles/Dockerfile.openshift
jonespm: Okay thanks for the catch.

jonespm:
I also noticed that default-libmysqlclient-dev was no longer needed here.

PR #358, opened 2024-11-15
I349 react upgrade

pushyamig:
I have updated to most of the package to recent except of couple

`@tanstack/react-query`: This version latest is version 5 and it has some change which need adding UseEffect which I want to avoid, but instead I upgraded to V4 that itself had major changes. The repo Owner changed npm package. V5 will be addressed in next release
https://tanstack.com/query/v4/docs/framework/react/guides/migrating-to-react-query-4

ESLint and its supported packages are one version behind the latest release. I chose not to update this since Eslint V9 has breaking changes `eslintrc.js` (eslint configuration) is changed to [eslint.config.js ] (https://eslint.org/docs/latest/use/configure/configuration-files) and the way it is configured within. So felt we can wait. 
https://eslint.org/version-support/

Material UI is V6, opted to stay on V5. Since most of our app still on V5

Use the cli command to check what is latest
1. ncu
2. npm show eslint@8 version

Testplan:
1. Do a docker clean and start the tool.
3. Make sure the Get/PUT function while adding and removing the tool works. Since all those API call from FE is handled by this @tanstack/react-query upgrade
4. Make sure No linting errors in the VSCode.


pushyamig:
@zqian and @jonespm this PR needs a Review with FE dependency update

frontend/app/components/Home.tsx
```diff
@@ -105,7 +105,7 @@ function Home (props: HomeProps) {
         </div>
       </MainContainer>
       <Typography component='footer' sx={{ textAlign: 'center' }}>
-        Copyright © 2022 The Regents of the University of Michigan
+        Copyright © {new Date().getFullYear()} The Regents of the University of Michigan
```
jonespm, line 108: I think it's good to update this but I feel like it should remain a static number. I've read debate out there both ways, but this way would pull from the client date which can be changed. I don't know if that matters. If we wanted to put this in backend home.html template that would be different. Do we have any existing apps with dynamic updates?

I'd probably just update it to 2024 for now and create a new issue to investigate this for best practice.

pushyamig: OK, I will make it static for now.

pushyamig: Created an issue https://github.com/tl-its-umich-edu/instructor-productivity-tools/issues/360

jonespm (approved):
Builds and starts up fine. Looks good after quick tests.

pushyamig:
Thanks @jonespm for the review, I will merge it now

PR #373, opened 2024-12-02
Apply workaround for oauth2 migration issue

### LINKED ISSUE #371: Apply workaround for oauth2 migration issue
I'd filed this issue but it currently isn't fixed and we still see a warning
https://github.com/Harvard-University-iCommons/django-canvas-oauth/issues/3

Two options are suggested, ignored all migrations from this 
```
MIGRATION_MODULES = {
    'canvas_oauth': None,  # This disables migrations for the canvas_oauth package
}
```

Or add this migration locally. I think I'm going to choose the option to add it locally.

jonespm:
Fixes #371


pushyamig:
I tested this and it seems to be working. These are the testing scenarios I followed.

1. The migration are added under the `app` as canvas_oauth and the error is gone while starting the app(django_migrations table)
2. I was able to generate a Oauth workflow for a new user other than me


I felt these scenarios should Be ok

pushyamig (approved):
Nice work, this workaround is though out very quickly and is reasonable solution given the fact we did not have to fork the repo.

jonespm:
Yeah, the only downside is that if this one is never applied upstream we'll either have to reconcile these or manually maintain these locally. The version is locked in the requirements.txt, I think I'll add a comment there about that to check the migrations if this is ever updated.

PR #380, opened 2024-12-05
Fix code scanning alert no. 1: Information exposure through an exception

jonespm:
Fixes [https://github.com/tl-its-umich-edu/instructor-tools/security/code-scanning/1](https://github.com/tl-its-umich-edu/instructor-tools/security/code-scanning/1)

To fix the problem, we should modify the `lti_error` function to log the detailed error message on the server and return a generic error message to the end user. This approach ensures that sensitive information is not exposed while still allowing developers to access the detailed error logs for debugging purposes.

1. Modify the `lti_error` function to log the detailed error message.
2. Return a generic error message in the JSON response.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/backend/canvas_app_explorer/lti1p3.py
+++ b/backend/canvas_app_explorer/lti1p3.py
@@ -66,7 +66,7 @@ def lti_error(error_message: Any) -> JsonResponse:
     :return: JsonResponse, with status 500
     """
     logger.error(f'LTI error: {error_message}')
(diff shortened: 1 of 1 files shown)


jonespm:
It was indicated that this error will not contain sensitive information and it makes resolving this issue faster if the user can provide a more specific issue.

PR #401, opened 2025-03-06
issue_386 updated the Django version to 4.2.20

zqian:


zqian:
How about moving it to 4.2.20, same the version in MyLA tool?

jonespm:
Yeah, dependabot should go all the way up but is the GH Action is failing for some reason. Was hoping that updating this config would get it going.

zqian:
OK, moved up to Django 4.2.20

zqian:
close the PR since the upgrade was done in #404

PR #403, opened 2025-03-07
Reconfigure to dependabot ignore major updates

jonespm:


.github/dependabot.yml
```diff
@@ -13,9 +13,8 @@ updates:
     allow:
       - dependency-type: "direct"
     ignore:
-      - dependency-name: "django"
```
zqian, line 16: Should we keep the django dependency rule, to be consistent with other T&L apps, like in https://github.com/tl-its-umich-edu/remote-office-hours-queue/blob/master/.github/dependabot.yml

jonespm: We're planning on also changing ROHQ to use this for backend and frontend https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/616. Seems worthwhile to hold everything back and only do major updates manually.

zqian (approved):

PR #412, opened 2025-04-01
issue_402 added field launch_url

zqian:
- added launch_url field to lti tool table
- the launch_url field is not null for non_LTI tools
- display "LAUNCH" button for non_lti tools


backend/canvas_app_explorer/models.py
```diff
@@ -40,6 +40,7 @@ class LtiTool(models.Model):
     support_resources = HTMLField()
     canvas_placement = models.ManyToManyField(CanvasPlacement)
     internal_notes = HTMLField(blank=True, null=True, help_text="a place to put helpful info for admins, not visible to users")
+    launch_url = models.CharField(max_length=255, blank=True, null=True)
```
jonespm, line 43: Not sure if we'll have to have long URL's but should _probably_ match the URL Length limit of 2048

backend/canvas_app_explorer/views.py
```diff
@@ -36,7 +38,8 @@ def list(self, request: Request) -> Response:
 
         logger.debug('available_tools: ' + ', '.join([tool.__str__() for tool in available_tools]))
         available_tool_ids = [t.id for t in available_tools]
-        queryset = models.LtiTool.objects.filter(canvas_id__isnull=False, canvas_id__in=available_tool_ids)\
+        queryset = models.LtiTool.objects.filter(Q(canvas_id__isnull=False, canvas_id__in=available_tool_ids)#).order_by('name')
```
jonespm, line 41: Should remove this unused comment. 

Nitpick: The new `.order_by` could probably also be better aligned or put on the preceding line if it fits.

backend/canvas_app_explorer/models.py
zqian: updated to 2028

backend/canvas_app_explorer/views.py
zqian: Done!

frontend/app/components/ToolCard.tsx
```diff
@@ -128,18 +128,30 @@ export default function ToolCard (props: ToolCardProps) {
...
-                  disabled={updateToolNavLoading}
-                  onClick={() => doUpdateToolNav({ canvasToolId: tool.canvas_id, navEnabled: false })}
-                />
-              )
-              : (
-                <AddToolButton
-                  disabled={updateToolNavLoading}
-                  onClick={() => doUpdateToolNav({ canvasToolId: tool.canvas_id, navEnabled: true })}
-                />
+                <Button
```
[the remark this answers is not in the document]
zqian: Done!

zqian:
@jonespm I have addressed all comments. Please review again! Thanks!

backend/canvas_app_explorer/models.py
jonespm: This also needs to be updated in the model here.

Also looking at this we should add help_text which would display for the admin. I'm trying to remember if the idea was that either the canvas_id OR the launch_url be added? What if there's values in both?

backend/canvas_app_explorer/admin.py
```diff
@@ -14,7 +14,8 @@ class LtiToolAdmin(admin.ModelAdmin):
         'privacy_agreement',
         'support_resources',
         'canvas_placement',
-        'internal_notes'
+        'internal_notes',
+        'launch_url',
```
jonespm, line 18: `launch_url` kind of gets lost down at the end of the form. The tuple order controls the form order.

I'd consider moving it up the list, either in the same line as `canvas_id` (like how `logo_image` is grouped) or immediately after `canvas_id`. If the `help_text` was defined on both or one of these fields it would make it clearer how to define this for an admin.

backend/canvas_app_explorer/models.py
jonespm: In testing, if `launch_url` is defined it doesn't matter what value is in `canvas_id`

So maybe help_text for launch_url should be something like.

"A link that will directly be launched by clicking on this card. If this is value is set then canvas_id is ignored"

zqian: help_text added!

backend/canvas_app_explorer/admin.py
zqian: moved the help_text up, next to canvas_id

jonespm:
Okay I think this looks better than works as described.

image

jonespm (approved):
Tested this and works as expected.

PR #414, opened 2025-04-04
413 Fix permission denied from supervisor.

### LINKED ISSUE #413: Errors starting up app on Openshift
When starting on Openshift we're seeing this error

PermissionError: [Errno 13] Permission denied: '/code/supervisord.log'

I believe that it's still trying to write a logfile and there's probably a config missing here from the change on #331.

Also it seems like after fixing this the page is blank and the static files aren't being collected as they previously were.

jonespm:
Fixes #413 Adding additional logfile config to the main block of supervisor


pushyamig:
The build is working on the IT-test instance.

pushyamig (approved):

Issue #240, opened 2022-07-07
Add some local event tracking


jonespm:
It might be possible to do the event tracking all through GA in #239 but if we'd like to do additional event tracking, for instance when a user adds/removes a tool from their course site we _may_ want to have that in a local database. I'd implement #239 first and see if sending a few custom events to GA seems good. Really what we want is to get an idea of usage and which tools are the most popularly added. 

In past projects we've used https://github.com/pinax/pinax-eventlog but there may be something better available. 

We've also used drf-api-tracking in some other projects so we probably should use that here too.

jaydonkrooss:
QA test plan: complete the first 2 steps of the [google analytics QA plan](https://github.com/tl-its-umich-edu/instructor-tools/issues/438#issuecomment-3145346679).

After interacting with the cards three different API  events will be tracked: show tools, enable, disable. These events will be viewable in the django admin page under API tracking

PR #426, opened 2025-06-09
issue #240 added eventlog for button clicks

zqian:


jonespm:
Since this uses DRF did you check to see if just adding https://pypi.org/project/drf-api-tracking/ `LoggingMixin` to the `LTIToolViewSet`  would work? Do we need a new event called on this from the frontend that is separate from the update in that class?

PR #488, opened 2025-11-13
#487 support asgi protocol

### LINKED ISSUE #487: Support async (ASGI) in app
Add support for Uvicorn worker along with Gunicorn to support async process. We will be supporting concurrency when fetching API response, and I think it work well have a Uvicorn worker. In addition we might need to switch to `servestatic`  for serving static files

The approach is similar to CCM effort https://github.com/tl-its-umich-edu/canvas-course-manager-next/issues/542

https://github.com/tl-its-umich-edu/canvas-course-manager-next/pull/544/files

-----------------------
Test Plan
1. In Development ( which is default), do a docker build remove old artifacts, make sure statics assets like `main-*.js` `main-*.css` files are server and webpack reload of image asserts works as well. VScode Debugging should work the same.
2. Test Prod mode locall, **remove** (comment) DEBUGPY_ENABLE=true  from `.env`and start the server. This will load the server in prod mode

For both cases Tool launches and enable/disable LTI tools.

pushyamig:
Fixes #487 

Test plan written in the issue, Local uses pure Uvicorn and Prod use a mix of Guincorn + Uvicorn worker. This design is something followed for CCM, ROHQ.

### DIFF
--- a/backend/asgi.py
+++ b/backend/asgi.py
@@ -11,6 +11,10 @@ import os
 
 from django.core.asgi import get_asgi_application
 
+from backend.debugpy import check_and_enable_debugpy
+
 os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'backend.settings')
+
+check_and_enable_debugpy()
 
 application = get_asgi_application()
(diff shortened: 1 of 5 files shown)


jaydonkrooss (approved):
changes look good. local testing passes

PR #501, opened 2025-11-21
Create an action that will attempt to build PR's

### LINKED ISSUE #500: Add an action to build ever PR to ensure that at least it doesn't cause any issue with the build
We recently had a PR merged to master fail because it didn't build.

We should minimally have an action that attempts to build the PR and fails when it does.

I can add this action in there.

jonespm:
Fixes #500


jonespm:
Yeah, I was thinking the same. Just having this as a separate file from the build/release made it so that this would only run this simple build task.

PR #505, opened 2025-11-25
Import only what is needed - Tree shaking

pushyamig:
This change improves the JS bundle size and fixes the immediate issue with size problem

```
instructor_tools        | asset main-dc831b456875e207a598.js 11.3 MiB [emitted] [immutable] (name: main)
instructor_tools        | asset main-dc831b456875e207a598.css 1.39 KiB [emitted] [immutable] (name: main)
instructor_tools        | Entrypoint main 11.3 MiB = main-dc831b456875e207a598.css 1.39 KiB main-dc831b456875e207a598.js 11.3 MiB
```

We can think about this webpack Optimization #504 later as well

### DIFF
--- a/frontend/app/components/AltTextHome.tsx
+++ b/frontend/app/components/AltTextHome.tsx
@@ -2,7 +2,7 @@ import { Box, Button } from '@mui/material';
 import React, { useState } from 'react';
 import { updateAltTextStartSync } from '../api';
 import ErrorsDisplay from './ErrorsDisplay';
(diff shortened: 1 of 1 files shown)


pushyamig:
Do a docker clean, removing all the docker assets, build and see the size reduced to small like ~12MB. 
for testing go back to `git checkout f68b31096413a7cb9f8c0594486df002d1dd2648` and see if the JS size is similar as well.

PR #531, opened 2026-01-02
issue_530 check for empty quiz question; do not update image list if …

zqian:
…there is no image returned from context


backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -207,13 +207,15 @@ def get_assignments(course: Course):
...
-            images_from_assignments.append({
-                'id': assignment.id, 
-                'name': assignment.name, 
-                'images': extract_images_from_html(assignment.description), 
-                'type': 'assignment',
-                'content_parent_id': None
-                })
+            assignment_images = extract_images_from_html(assignment.description)
+            if len(assignment_images) > 0:
+                images_from_assignments.append({
```
pushyamig: This change is ok, I feel there is repeat for this append. 

So simple suggestion would be extract the append to a different function and pass needed items then
```
if len(assignment_images) > 0:
    append_content_items(type=assignment, content_item)
```

Issue #465, opened 2025-10-15
Convert to AzureOpenAI and send images


jonespm:
The current implementation uses google gemini to send images, we'll need to take the code he has and convert it to AzureOpenAI likely based on the examples from 

https://github.com/umich-ai-resources/ITS-Examples/blob/main/Python/python-vision-local-file.py

- [ ] I think we need to have a Nice Prompt that fetched the Alt text. Current Canvas Flags if alt text is > 120 character. Need some guideline what is the Optimal characters 
- [ ] Currently with AzureOpenAI a single call to fetch alt text takes about ~3sec, so we need to implement Async call when making `x` number of call. Looking at the AzureOpenAPI library, they are using `httpx` which supports async. Need to investigate whether the Azure client supports Multiple simultaneous call or we have to do it with `asyncio`

The below code basic wiring with AzureOpenAI

```
def get_alt_text_from_openai(imagedata, mime_type):
    start_time: float = time.perf_counter()
    client = AzureOpenAI(
        api_key=os.getenv('OPENAI_API_KEY'),  
        api_version=os.getenv('API_VERSION'),
        azure_endpoint = os.getenv('OPENAI_API_BASE'),
        organization = os.getenv('OPENAI_ORGANIZATION'))
    
    prompt = """
        As an AI tool specialized in image recognition, generate concise and descriptive alt text for this image. 
        The description should be suitable for a student with a 
        vision impairment taking a quiz. Do not include phrases 
        like 'This is an image of...'. Provide only one concise 
        option with no further explanation.
        """
    messages=[
            {"role": "system", "content": prompt},
            {"role": "user", "content": [
                {"type": "image_url", "image_url": {
                    "url": f"data:image/jpeg;base64,{imagedata}"}}
            ]}
        ]

    response = client.chat.completions.create(
        model=os.environ['MODEL'],
        messages=messages,
        temperature=0.0,
    )
    end_time: float = time.perf_counter()
    logging.info(f"OpenAI call duration: {end_time - start_time:.2f} seconds")
    alt_text = response.choices[0].message.content
    logging.info(f"OpenAI response: {alt_text}")
    return alt_text
```

PR #533, opened 2026-01-05
#532 moving the image download call to Background task

### LINKED ISSUE #532: Move the image content download to Background task
After the #517 merge we need to move downloading the image content to background task and also base 64 encoded.

pushyamig:
Fixes #532 

1. When the images content download process has error, a custom exception is raised and scan is marked as failed. 
2. So AI call will be next happens after the image download of the content. ( A**I call is still in place but in a very rudimentary state so the issue #465 finish the design)**
3. Repurposed the `GetContentImages` from #517 , so the Image Download API logic is preserved. The PR just handles how to handle the exception and combined object with `image id, url, base64 encoded`
    1. `GetContentImages: images_object` holds all the content type info 
    2. we flatten the object and just pull the images object from that. Unit test`test_flatten_images_from_content.py` has both samples types object that are send to Image download API

This object can be fed to the AI call, and AI call needs to return the alt text and `content`  can be dropped when storing the final return.  where the AI call is made the return object to background scan file needs a change, but as of now, i will leave that the issue #465 handle
 ```
    [
            {
                'image_id': 44125878,
                'image_url': 'https://canvas-test.it.umich.edu/files/44125878/download?verifier=fPOPQAIAmmu8numuuCQRwLG93ORQWs0w9yFMs8QZ&download_frd=1',
               'content': <base64_encode_image>
            },
          ]
```

### DIFF
--- a/backend/canvas_app_explorer/alt_text_helper/get_content_images.py
+++ b/backend/canvas_app_explorer/alt_text_helper/get_content_images.py
@@ -74,7 +74,7 @@ class GetContentImages:
         try:
             start_time = time.perf_counter()
             images_list = self.flatten_images_from_content()
-            logger.info(f"Image List : {images_list}")
+            logger.debug(f"Image List : {images_list}")
             
             logger.info(f"Retrieved {len(images_list)} images for course_id: {self.course_id}")
             images_content = self.get_image_content_from_canvas(images_list)

### CODE REVIEW

File: backend/tests/test_flatten_images_from_content.py [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+            },
+            {
+                'image_id': 44125882,
+                'image_url': 'https://canvas-test.it.umich.edu/files/44125882/download?verifier=n0zb6AyIuBlbgwskCO2p18f4Dx7pQmqtZ6KirTcW&download_frd=1'
+            },
+            {
+                'image_id': 44125883,
+                'image_url': 'https://canvas-test.it.umich.edu/files/44125883/download?verifier=mGiQl2PLgJGm1gkhhXNzDHzWligy5SkUEXamICAn&download_frd=1'
+            },
+        ]
```
pushyamig: This is what fed to the Canvas API for fetching the image content ( or downloading the content)

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -113,6 +115,15 @@ def _has_fetch_error(result) -> bool:
...
+    try:
+        # what to do with this content images depending on the AI call handled so the exception handled there as well.    
+        images_by_course = content_images.get_images_by_course()
+    except ImageContentExtractionException as e:
+        logger.error(f"Error extracting image content for course {course.id}: {e}")
+        update_course_scan(course.id, CourseScanStatus.FAILED.value)
+        return
+
     # DB call to persist ContentItem and ImageItem records
     save_scan_results(course.id, filtered_content_with_images)
```
pushyamig: I believe the AI Call will happen before the `save_scan_results`.

jonespm (approved):
I think this looks good and aligns with the current workflow.

As a nitpick, I know this isn't part of this change but the code in `fetch_and_scan_course ` that generates a request from seems like it should be moved into a separate method. Where we can just have something like

`canvas_api: Canvas = get_canvas_api_from_request(request)`

This utility method seems like it could be useful to have outside of this.

backend/canvas_app_explorer/alt_text_helper/get_content_images.py
```diff
@@ -68,34 +67,69 @@ def get_alt_text_from_openai(self, imagedata):
...
+
+        Optionally, a flattened `images_input` list may be provided. If not provided,
+        the instance's `images_object` will be used.
         """
         try:
             start_time = time.perf_counter()
-            images = ImageItem.objects.filter(course_id=self.course_id).values('image_id', 'image_url')
-            images_list = list(images)
+            images_list = self.flatten_images_from_content()
+            logger.info(f"Image List : {images_list}")
```
zqian: Should the log level be debug instead?

backend/tests/test_get_content_images_exceptions.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class DummyGetContentImages(GetContentImages):
+    def __init__(self, images_object):
+        # don't need a real canvas_api for this test
+        self.course_id = 1
+        self.canvas_api = None
+        self.images_object = images_object
+
+    def get_image_content_from_canvas(self, images_list):
```
zqian: maybe the function should be renamed to reflect the failed return result?

pushyamig:
Thanks for the review. I am merging it now.

PR #540, opened 2026-01-06
#539 removing the course id from FE api request

### LINKED ISSUE #539: Frontend API call shouldn't send course id as a query parameter
The alt text helper url/view send the course_id as part of the Http request. This is not needed instead should pull the course id from the request session which is something existing LTI tool Http requests follows, So we are adopting alt text helper in the same format

pushyamig:
Fixes #539 

The Course scan should pull the course id from the request session as it is stored as part from the LTI launch.

Testing:
Make sure the course scan works from multiple courses.


pushyamig:
I am merging it!

Issue #538, opened 2026-01-06
Handle Canvas Course using images citing from external sources


pushyamig:
This issue is found in this course where this [page](https://canvas-test.it.umich.edu/courses/562259/pages/aggression) has an image sources cited from external website and background scan doesn't handle this. **Scan made an assumption that all images are uploading from canvas**

```
[canvas_tools_alt_text_scan.py:378] Processing img tag: <img alt="" class="media-object noba-author pull-right" height="150" loading="lazy" src="https://nobaproject.com/images/shared/author_photos/000/000/019/large.jpg" width="150"/>
[ERROR] [canvas_tools_alt_text_scan.py:368] Error parsing img src URL 'https://nobaproject.com/images/shared/author_photos/000/000/019/large.jpg': File ID not found in URL path
```

What this mean is we need to have a column in the image type saying `Canvas` or not and since these URL seems to be public should allow download of images, but the scan process now need to differentiate image URL from canvas or public and should handle download image content and Base64 encoding

pushyamig:
I feel for public URL AI can handle to download the image and provide an alt-text if we provide a link to it when making a request, but regardless, there should be differentiator for public vs canvas images.

jonespm:
I believe `get_image_content_sync `is assuming that there's a file_id currently. If there isn't (because it's external) it can just try to download the images directly with requests rather than using CanvasAPI. I think this will be good to implement but doesn't need to be done highest priority right now.

jonespm:
It was commented on #547 that we may be able to just drop `image_id` as part of these changes.

PR #542, opened 2026-01-08
#541 pulling content image info to UI

### LINKED ISSUE #541: Basic wiring of fetching the content images and presenting in the Ui
This is a bare minimum approach where we pull the content image ( pages, assignments, quizzes) from DB and show it in the UI. 

We Gonna pull all the images at once, this might be a temporary state of the pull might be refined more as we go

pushyamig:
Fixes #541

This is not Beautiful looking but functional in a basic way. The PR just pull all the images associated with a Content_type ( assignment, page, quiz ( also include quiz_questions) and paginates. Pagination is not part of the design so need to adjust based on the Figma version.

### DIFF
--- a/backend/canvas_app_explorer/alt_text_helper/views.py
+++ b/backend/canvas_app_explorer/alt_text_helper/views.py
@@ -14,8 +14,6 @@ from backend import settings
 from backend.canvas_app_explorer.canvas_lti_manager.django_factory import DjangoCourseLtiManagerFactory
 from backend.canvas_app_explorer.models import CourseScan, CourseScanStatus
 from backend.canvas_app_explorer.serializers import ContentQuerySerializer
-import re
-from urllib.parse import urlparse, parse_qs, urlencode, urlunparse
 
 logger = logging.getLogger(__name__)
 


.env.sample
```diff
@@ -30,7 +30,9 @@
 # A comma separated list of domains where you want to allow the application to be framed
 # CSP_FRAME_ANCESTORS=umich.instructure.com,*.loophole.site,*.umich.edu,umich.edu,*.ngrok-free.app,*.instructure.com
 # CSP_CONNECT_SRC=*.google-analytics.com,umich.edu,*.googletagmanager.com,
-# CSP_IMG_SRC=umich.edu
+# 'umich.edu,https://canvas-test.it.umich.edu,https://*.canvas-user-content.com,https://inst-fs-iad-test.inscloudgate.net,https://cdn.inst-fs-iad-test.inscloudgate.net'
+# since we are allowing public image urls, we need to allow https:. Canvas seems to use various domains for images. 
+# CSP_IMG_SRC=https:
```
pushyamig: So for testing we are allowing all image downloads via `https:` But One thing i noticed is since we are providing the canvas Preview url ( might also applied to external url)  image loads is slow.

This kind of let's me think how canvas external URL?

backend/canvas_app_explorer/models.py
```diff
@@ -141,7 +141,7 @@ class ImageItem(models.Model):
         db_column='content_id',
         related_name='images',
     )
-    image_id = models.BigIntegerField()
+    image_id = models.BigIntegerField(null=True, blank=True)
```
pushyamig: Since we have external images, so we won't have `image_id` only `image_url` and we need to allow null so more on this will be addressed #538 . I had a failing test so I need to handle this now instead of in that #538

frontend/app/interfaces.ts
```diff
@@ -78,4 +78,19 @@ interface AltTextLastScanDetail {
...
+  image_alt_text: string | null
+}
+
+interface ContentItemResponse {
+  content_id: number
+  content_name: string
+  content_parent_id: number | null
+  content_type: string
+  images: ContentImage[]
+}
```
pushyamig: This is the Content object that is returned from the JSON response

PR #545, opened 2026-01-09
remove the test not needed

### LINKED ISSUE #541: Basic wiring of fetching the content images and presenting in the Ui

pushyamig:
Fixes #541
#542 This issue was caused because i did not delete the test when I removed the function (https://github.com/tl-its-umich-edu/instructor-tools/pull/542/changes/ef5660bc62b13e312b89eb89b33c3df5bb6e57cf)


zqian (approved):
The removal of the test case makes sense.

pushyamig:
This PR has this change #546 , so this PR is not needed anymore. I wasn't sure which one would get the review. so I opened as separate PR. 

Closing this PR

PR #550, opened 2026-01-13
I482 prod mode qworker

### LINKED ISSUE #482: Qworker: Enable production deployment settings
Add more configurations and settings changes to this issue that are needed before deployment.

From initial alt text setup: "Create an issue qworker support prod run without `watchfiles` as it reload code when changes"

_Originally posted by @pushyamig in https://github.com/tl-its-umich-edu/instructor-tools/pull/478#discussion_r2514773308_

pushyamig:
Fixes #482 

Interestingly qworker doesn't need reload separate with `watchfiles`. This was taken from CCM and i don't know what is so different between the setup that this app doesn't needs it

How I tested this change..
1. Do a `docker compose build --no-cache`
2. Make changes to the background task file `canvas_tools_alt_text_scan.py` and `get_content_images.py` and see these changes are picked up when running the changes
3. Added `Blessed` pkg just see the qworker stats in the container commandline. We don't rely on this but this is useful to quickly check if the qworker is up and running or not, how many are started

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,3 +45,18 @@ All test are in the `tests` folder. To run the tests
 ```sh 
 docker exec -it instructor_tools python manage.py test
 ```
+### Django Queue
+1. Getting the Alt text from course images run as a background task.
+2. We are using [Django ORM](https://django-q2.readthedocs.io/en/master/brokers.html#django-orm) is set a default message Broker.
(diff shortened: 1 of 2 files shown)


pushyamig:
I Resolved conflicts and upcoming changes from main

Issue #535, opened 2026-01-06
Move Canvas API creation from fetch_and_scan_course into shared utility


pushyamig:
Commented by @jonespm 

https://github.com/tl-its-umich-edu/instructor-tools/pull/533#pullrequestreview-3631769583

PR #559, opened 2026-01-20
I538 add external url support

### LINKED ISSUE #538: Handle Canvas Course using images citing from external sources

pushyamig:
Fixes #538 #535

1. Now content can have internal and external images and scan can process them. See Course (403334) for some examples
2. Model change, removed the `image_id` which is a canvas thing and external images doesn't have them. That id is redundant and not needed for canvas download since all canvas cares is URL
3. The code is attaching the Bearer token only for Canvas domain, and not for external images
4. Fixed the tests, did not write new tests

**How to add External image?**
 1. Copy a external URL of the image something like https://nobaproject.com/images/shared/images/000/000/099/original.png and go to the Content description text box and simply paste. This way allows to save image **not** as decorative. This is a bit of nuance to consider when setting up the site. Since we ignore Decorative image in the scan. so you might feel there is something wrong if these images don't show up in `ImageItem` table

2. Make sure to set `CSP_IMG_SRC='https:'` in the `env`. This will allows for see the external images from UI without CSP error. I think this is going to be default for this phase of the project. Think about Dynamic CSP later phase if it is worth it.

### DIFF
--- a/backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
+++ b/backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
@@ -2,7 +2,7 @@ import asyncio
 import logging
 from django.db import transaction
 from urllib.parse import urlparse, parse_qs, urlencode
-from typing import List, Dict, Any, Optional, Tuple, TypeVar, Callable, Union
+from typing import List, Dict, Any, Optional, TypeVar, Callable, Union
 from asgiref.sync import async_to_sync
 from django.test import RequestFactory
 from django.contrib.auth.models import User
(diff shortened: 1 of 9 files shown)


zqian (approved):
Verified on localhost + Canvas test. Images with External urls can be downloaded successfully; All test cases pass.

PR #571, opened 2026-01-27
issue_482 added configuration variable RUN_QWORKER_DEV_MODE

zqian:
- added RUN_QWORKER_DEV_MODE to .env.sample
- default value is true if not set

To test:
- on localhost, do a `docker compose build --no-cache` and `docker compose up`: search for "qworker: Running in DEV mode" line in the log 
- on OpenShift server, search for "qworker: Running in DEV mode" line in the log


pushyamig:
Can this PR be updated to the latest?

pushyamig:
I can see the changes with if/else picking up the right stuff

PR #579, opened 2026-01-28
Expose CourseScan to Django Admin

### LINKED ISSUE #574: Expose CourseScan to Django Admin
This is Quick workaround where we can quickly mark `state=failure` incase of long running loop.

jonespm:
Also refactor so that the CourseScanStatus is a choices on the model and can be used in the admin.

Fixes #574


pushyamig:
Reviewing now

PR #580, opened 2026-01-28
issue #565 Some worker exceptions don't mark scan as FAILED, causing get_last_scan to poll indefinitely

### LINKED ISSUE #565: Some worker exceptions don't mark scan as FAILED, causing get_last_scan to poll indefinitely
Problem:

When the background worker task `fetch_and_scan_course` encounters an exception, the `CourseScan` status is not updated to `[FAILED]`. This causes the frontend's `get_last_scan` polling to hang indefinitely, as it's waiting for a status change from pending/running to either completed or failed.

Some things are checked but there's still failures that come up (mostly in development) that can cause this to hang.

These 3 were likely causes

* Content fetch failures - When `unpack_and_store_content_images()` returns False (indicating a fetch error), the function logged the error but returned without updating the status to `FAILED`
* Async image gathering exceptions - Any exception during `async_to_sync(get_courses_images)(course)` (from assignments, pages, quizzes fetching) that wasn't caught by individual functions would crash the worker silently
* Course object creation exceptions - The line `course: Course = Course(canvas_api._Canvas__requester, {'id': course_id})`] could throw an exception that went uncaught

I'm not sure how to test this but it's easy to create this situation locally by just breaking one of these locally.

jonespm:
Fixes #565

### DIFF
--- a/backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
+++ b/backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
@@ -44,7 +44,7 @@ def fetch_and_scan_course(task: Dict[str, Any]):
     course_id = int(task.get('course_id'))
 
     # adding a status before start of the scan, if this DB action failed no need to stop next steps of fetching content images
-    update_course_scan(course_id, CourseScanStatus.RUNNING.value)
+    update_course_scan(course_id, CourseScanStatus.RUNNING)
 
     try:
         # Fetch course content using the manager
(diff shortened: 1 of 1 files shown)


backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -46,45 +46,53 @@ def fetch_and_scan_course(task: Dict[str, Any]):
...
+        except ImageContentExtractionException as e:
+            logger.error(
+                f"ImageContentExtractionException while processing alt text for course_id {course_id}: {e}",
+                exc_info=True
+            )
+            update_course_scan(course_id, CourseScanStatus.FAILED.value)
+            return
+
+        # Update that the course scan is completed
+        update_course_scan(course_id, CourseScanStatus.COMPLETED.value)
```
zqian: I feel the logic could be simplified as:

- in line 73 check for 'if state_of_content_fetch' boolean instead
- move line 89 be moved after line 79, since this is only return of "CourseScanStatus.COMPLETED.value"
- return "CourseScanStatus.FAILED.value" for all other cases

PR #582, opened 2026-01-28
Added back in the missing log timer and make decorator accept async

### LINKED ISSUE #581: Update log_execution_time decorator to make it work in async or sync contexts
The decorator log_execution_time only works currently on sync methods. I had some code to make it detect and work in both async or sync I think that should be retained. I believe it was lost somehow from the change on #564

Additionally the _worker_async method lost this decorator since it couldn't handle async.

jonespm:
Had to do some changes to make the decorator accept sync or async calls.

Fixes #581 

This isn't the highest priority to get into this release but it should work out fine. I'd added this as part of another issue but for some reason the commit was on the wrong branch.

### DIFF
--- a/backend/canvas_app_explorer/decorators.py
+++ b/backend/canvas_app_explorer/decorators.py
@@ -28,7 +28,7 @@ def log_execution_time(func: Callable) -> Callable:
             finally:
                 end_time: float = time.perf_counter()
                 duration = end_time - start_time
-                logger.info(f"{func.__name__} execution time: {duration:.2f} seconds")
+                logger.info(f"{func.__name__} (async) execution time: {duration:.2f} seconds")
         return async_wrapper
     else:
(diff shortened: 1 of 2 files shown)


backend/canvas_app_explorer/decorators.py
```diff
@@ -8,23 +9,36 @@
...
+        @wraps(func)
+        async def async_wrapper(*args: Any, **kwargs: Any) -> Any:
+            start_time: float = time.perf_counter()
+            try:
+                result = await func(*args, **kwargs)
+                return result
+            finally:
+                end_time: float = time.perf_counter()
+                duration = end_time - start_time
+                logger.info(f"{func.__name__} execution time: {duration:.2f} seconds")
```
zqian: Would it be helpful to add "sync" vs "async" text into the log message?

Issue #586, opened 2026-01-29
extract_images_from_html logs an empty message when no images found.


jonespm:
When no images are found the method `extract_images_from_html` seems to log an empty array which adds unnecessary noise to the logs. This should either be reduced to debug or just not log when it's empty. 

https://github.com/tl-its-umich-edu/instructor-tools/blob/c36994af9be086a201028ed0a2f82654b49a5db1/backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py#L429

PR #587, opened 2026-01-29
#583 allow canvas public urls

### LINKED ISSUE #583: Canvas allow Public URL  and User images in course without /course/ context in the URL
- Canvas Public URL's  seems to be pulled into Courses https://github.com/instructure/canvas-lms/tree/master/public/images
  - https://canvas-test.it.umich.edu/images/block_editor/section-resources.png
- Also User images seems to be accessible to course members (student primarily): https://canvas-test.it.umich.edu/users/5/files/55555/preview?verifier=CegEgfos9mXob5S2JttKTOeSet4Rx39n

With Canvas public URL's doesn't have files ID attached, so when we encounter them 
1. During the scan/update of alt-text call They don't need to be part of the URL rewriting we are doing. since these are public
2. It might be ok to send Bearer token with trying to get the image content, since they are going hit the canvas server

User Image: when you are trying to importing the images to assignment, pages, quizzes there is an option available apart from `Upload image` and `Course Images` there is `User Image` and this is the another way to pull in the images to the course. Interestingly Canvas when pulling the images doesn't download to course `Files` Tool but instead keep the `/user/` context.

The app needs to support this as well when scan and updating the alt-text back to canvas

pushyamig:
Fixes #583 #586

We still send Canvas token when fetching **canvas** file content as I felt, this token is going to canvas server should be safe!

### DIFF
--- a/backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
+++ b/backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
@@ -428,7 +428,8 @@ def extract_images_from_html(html_content: str) -> List[str]:
             download_url = img_src
         images_found.append(download_url)
     
-    if images_found: logger.info(images_found)
+    if images_found:
+        logger.info(images_found)
     return images_found
 
 # Helper function to append image items if images exist


backend/canvas_app_explorer/alt_text_helper/alt_text_update.py
```diff
@@ -526,7 +526,8 @@ def _transform_image_url(self, parsed) -> str | None:
                         urlencode(new_query, doseq=True),
                         parsed.fragment
                     ))
-            return None
+            else:
+                return parsed.geturl()
```
pushyamig: This again with updating the alt-text we missing file_id we just return the original URL

PR #588, opened 2026-01-29
feat: Add contextual linking for image titles in Alt Text Helper

jonespm:
Generate and display Canvas URLs as clickable links on image titles, enabling users to navigate back to source content. Includes theme-level link styling defaults and comprehensive test coverage.

@pushyamig @zqian I've tested with assignments, pages, quizzes and quiz questions.

It seems like the closest we can get on the questions is to the question_tab. So updating it to do that. If you see a better way we can do this. 

https://community.instructure.com/en/discussion/600774/if-i-have-a-course-id-quiz-id-and-a-question-id-can-i-create-a-url-to-directly-edit-the-question

### DIFF
--- a/backend/canvas_app_explorer/utils.py
+++ b/backend/canvas_app_explorer/utils.py
@@ -36,8 +36,8 @@ def generate_canvas_content_url(course_id: int, content_type: str, content_id: i
     elif content_type == 'quiz':
         return f'https://{canvas_domain}/courses/{course_id}/quizzes/{content_id}'
     elif content_type == 'quiz_question':
-        # Quiz question URLs point to the quiz with an anchor to the question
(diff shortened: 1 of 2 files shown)


frontend/app/components/ContentImageCard.tsx
```diff
@@ -93,9 +93,24 @@ export default function ContentImageCard({
...
+            href={contentImage.canvas_link_url}
+            target="_blank"
+            rel="noopener noreferrer"
+            fontWeight={600}
+            noWrap
+            sx={{
+              display: 'block',
+            }}
+          >
+            {contentImage.content_name || 'Untitled'}
```
zqian: Is it useful to include the content type here, like 'Untitled {contentImage.content_type}? 

I do not see the content type displayed once user is in the "Content Review" page:

PR #594, opened 2026-01-30
Add parent content name display for quiz questions

jonespm:
Display quiz questions with their parent quiz name as "Quiz Name : Question Name" in the alt text reviewer. Fetches parent names from existing queryset and updates frontend interfaces accordingly.

### DIFF
--- a/frontend/app/interfaces.ts
+++ b/frontend/app/interfaces.ts
@@ -87,7 +87,7 @@ interface ContentImage {
 
 interface ContentItem {
   content_id: number
-  content_name: string | null
+  content_name: string
   content_parent_id: number | null
   content_parent_name: string | null
   content_type: string
@@ -96,7 +96,7 @@ interface ContentItem {
 
 interface ContentImageEnriched extends ContentImage {
   content_id: number
-  content_name: string | null
+  content_name: string
   content_parent_id: number | null
   content_parent_name: string | null
   content_type: string


frontend/app/components/ContentImageCard.tsx
```diff
@@ -90,6 +90,14 @@ export default function ContentImageCard({
     return <StatusChip label="Not yet reviewed" size="small" />;
   };
 
+  const getContentTitle = () => {
+    return contentImage.content_parent_name 
+      ? `${contentImage.content_parent_name} : ${contentImage.content_name || 'Untitled'}` 
+      : (contentImage.content_name || 'Untitled');
```
zqian: Suggestions for missing names:

return contentImage.content_parent_name 
      ? `${contentImage.content_parent_name} : ${contentImage.content_name || 'Untitled quiz question'}` 
      : `Untitled quiz: contentImage.content_name || 'Untitled quiz question'`;

jonespm: I think in this suggestion frontend would need to make a type determination then and add it. 

Perhaps this application of an "Untitled" should just be done in the backend? 

I also don't think this is realistically likely since names are a required field for assignment, pages, quizzes. It seems to let you save a Quiz Question without a name but the UI auto applies "Question" to the name if blank. But perhaps a quick check in the backend would be easiest.

PR #598, opened 2026-02-02
#595 #583 #586 ignore copy paste links

### LINKED ISSUE #595: Copy/Paste image Link from other courses should be ignored
When you properly import (use `Copy To`) content ( page/assignment/quizzes) to a different course, canvas seems to download the image resources to the importing course, but instead is user use the Copy/paste Option that the image link will remain from that past resources, this is a wrong user action and can lead to not showing the resources to Student. 

If we see an URL course id is something different from current course, then simply ignore these image links otherwise this will cause failure to scan process

```
<img
    src="https://canvas-test.it.umich.edu/courses/177/files/51102/preview"
    alt="week 6 icon Song guanware gourd shaped crackle glazed vase.jpg"
    width="417"
    height="480"
    data-api-endpoint="https://canvas-test.it.umich.edu/api/v1/courses/177/files/51102"
    data-api-returntype="File"
    loading="lazy"
  >
```

pushyamig:
Fixes #595 #583 #586

This PR is fixing 3 things
1. Allow use of Canvas Public URL's #583 
2. Allow Course member User URL format #583
3. Ignore if image URL contains Course ID that is due to Copy/paste of description and this is lousy approach and course members ( student) doesn't have access to such files, so it is pointless for use to cater to this. #595 
4. Don't log when no images #586 

Please look at the issues for sample URL

### DIFF
--- a/backend/canvas_app_explorer/alt_text_helper/alt_text_update.py
+++ b/backend/canvas_app_explorer/alt_text_helper/alt_text_update.py
@@ -440,14 +440,22 @@ class AltTextUpdate:
             for image_payload in next(c for c in self.content_with_alt_text if c['content_id'] == content_id)['images']:
                 url_for_update = image_payload['image_url_for_update']
                 
+                # Skip if URL transformation failed (should not happen in normal flow)
+                if url_for_update is None:
+                    logger.warning(f"Skipping image with None url_for_update in content {content_id}")
+                    continue
+                
                 # Determine if url_for_update is a file_id (numeric string) or full URL
-                is_file_id = url_for_update and url_for_update.isdigit()
+                is_file_id = url_for_update.isdigit()
                 
                 matched = False
                 if is_file_id:
                     # For file_id, check if it's contained in the img src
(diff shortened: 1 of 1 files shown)


backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -359,25 +358,79 @@ def get_quiz_questions_sync(quiz: Quiz):
...
+            logger.debug(f"Including public Canvas image: {img_src}")
+            return True
+        
+        # User files (e.g., /users/{id}/files/{file_id}) - always include
+        if 'users' in parts and 'files' in parts:
+            logger.debug(f"Including user file: {img_src}")
+            return True
+        
+        # Look for /courses/{course_id}/ pattern and validate
+        if 'courses' in parts:
```
zqian: Looks like this /courses/ URL pattern is the only check needed, and we can remove the checks for 'images'  and 'users" above?

PR #600, opened 2026-02-02
#599 Ignore images tags without src

### LINKED ISSUE #599: Ignore Image tags without `src`
Currently we are not adding a check for an empty image src. Looks like this is possible to have `` Maybe this is a manual add, if instructor is building his own page.

pushyamig:
Fixes #599

### DIFF
--- a/backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
+++ b/backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
@@ -467,6 +467,7 @@ def extract_images_from_html(html_content: str, course_id: int) -> List[str]:
 
         # ignore images without src
         if not img_src:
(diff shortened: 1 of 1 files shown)


pushyamig:
I think the build is failing since Runner did not run this and probably things are timed out. This changes doesn't impact the when building

```

Internal server error. Correlation ID: 870db5d4-a56c-481d-b40c-9ea9dd294b2d
--
The job was not acquired by Runner of type hosted even after multiple attempts

```

pushyamig:
https://www.githubstatus.com/

```
Update - We continue to investigate failures impacting GitHub Actions hosted-runner jobs.
We have identified the root cause and are working with our upstream provider to mitigate.
This is also impacting GitHub features that rely on GitHub Actions (for example, Copilot Coding Agent and Dependabot).
Feb 02, 2026 - 21:13 UTC
```

Probably there is an outage so that's why the build is failing.

PR #630, opened 2026-02-27
#629 remove the paste plugin

pushyamig:
Fixes #629 

I don't feel the `paste` plugin was ever supported so I am removing it. https://github.com/jazzband/django-tinymce/tree/master/tinymce/static/tinymce/plugins

If you want to include it needs to be included as part of CDN, that's what docs says: `TINYMCE_JS_URL = 'http://debug.example.org/tiny_mce/tiny_mce_src.js'`

This is broken in prod and doesn't effect the functionality, so i am just removing what is not needed


jaydonkrooss (approved):
Error is gone during local testing. All other aspects of the text editor look good to me in the admin page.

jonespm (approved):

PR #641, opened 2026-03-18
#567 add Exit Dialog for Review page, back/refresh listeners

### LINKED ISSUE #567: Review UI Exit Dialog
Additional UI work from #495. 

Part of the review UI drafts included a warning for the user that exiting or going back would lose their progress. 

When any labels have been accepted and/or modified, show a pop-up modal if the user is about to go back or redirect. For example, 

Image

jaydonkrooss:
Uses MUI Dialog if any items are reviewed + unsubmitted and the user wants to leave or refresh.

Test cases for no actions taken => no dialog will appear
- during browser refresh 
- browser back button 
- Navigation - click "go back" on UI or click Instructor Tools breadcrumb

For all three above cases, if any actions are taken => dialog will appear 

The types of actions that will set up the exit guard:
- Approve
- Skip
- Mark as decorative
- Edit AI label (automatically approves label)

After submitting any changes, the exit guard will not appear upon navigation or clicking the "done" button.

Note: ~~I also added a skill to help with frontend testing via Chrome Dev tools MCP~~
I removed the dev tools skill; We may want to investigate other MCP's / CLI's to leverage for automated browser validations - better to explore in a separate issue, and probably after we get regular UI testing up an running


zqian (commented):
I found one change maybe unintentional: as soon as I made changes to the alt text area, the "Approve" button is highlighted even though I did not click on it. This is different from the current prod version.

zqian (approved):
Tested the Dialog window works as expected.

PR #660, opened 2026-04-13
#659 Multiple scan support for course

### LINKED ISSUE #659: Support multiple scan save for same course
Currently each scan a course, we delete prior remaining course content/images. But going forward we want to support and keep the each scan and it artifact separate

1. In case of multiple course scan, fetch and send the latest in DB
2. Don't delete the prior scan fetch
3. But when user try to do the adding alt-text to a image, we do delete the content.images. This remain same as before. But we always serve the latest from the scan to the use

Changes Made:
1. CourseScan `id` is FK with ContentItem as `course_scan_id`
2. ContentItem `id` is FK to ImageItem table as `content_item_id`
3. course_id is now only available from CourseScan and not unique anymore
4. The migration supports existing data to update to new schema without any manual changes.

This is Pre-req of the Report error, so I am making this issue as child to that

pushyamig:
Fixes #659 

Testing Locally: 

1. Try with a new DB and do a smoke test of workflow, see if multiple scan workflows holds well
2. Apply the migration to existing DB you have locally. Do the full range testing for this usecase
3. Pull the Data from a Non-prod DB, and see if these changes fair with existing QA data. I was almost tempted to test this on Prod by pulling down to locally, but halted **but tested the migration `app-explorer-Test-DB-instance`**. Steps how to pull the data from Non-prod or prod to you local [DB](https://docs.google.com/spreadsheets/d/1STBdddkQzNnOqR3hbPl3KWijHTsV_eSxaPQipYIJ4Pw/edit?gid=479353435#gid=479353435). **We can test these migration application for all 3 campuses Prod DB.  But i want to hear what our strategy should be for smooth transition.**

All the testing i did are [here](https://docs.google.com/spreadsheets/d/1STBdddkQzNnOqR3hbPl3KWijHTsV_eSxaPQipYIJ4Pw/edit?gid=0#gid=0)

Tips: 
1. Try new Build when testing this PR
2. swap DB by renaming the `.data` files that is locally mapped to the our disk
3. Try to revalidate/re-connect from Dbeaver or sql client, since there is some fuzziness i saw where FK-PK relationship appeared as not applied

What this PR is not?
1. Not returning total image count to course scan table. we will details with #622 
2. No new tables created to error reporting issue #622


backend/canvas_app_explorer/serializers.py
```diff
@@ -88,6 +88,7 @@ class ContentQuerySerializer(serializers.Serializer):
     content_type = serializers.ChoiceField(
         choices=ContentItem.CONTENT_TYPE_CHOICES
     )
+    course_scan_id = serializers.IntegerField(required=False)
```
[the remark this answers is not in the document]
pushyamig: Done as part of this https://github.com/tl-its-umich-edu/instructor-tools/pull/660/changes/fba49c053a091ca31d992b639e3f277c5a6c81f9

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -153,54 +155,54 @@ def _has_fetch_error(result) -> bool:
...
     except (DatabaseError, Exception) as e:
-        logger.error(f"Error updating CourseScan for course_id {course_id} to status {status.value}: {e}")
+        log_context = f"course_scan_id={course_scan_id}"
+        logger.error(f"Error updating CourseScan {log_context} to status {status.value}: {e}")
     
-def save_scan_results(course_id: int, items: List[Dict[str, Any]]):
+def save_scan_results(course_scan_id: int, course_id: int, items: List[Dict[str, Any]]):
     """
     Save the scan results into the database within a transaction.
     Deletes previous ContentItem and ImageItem records for the course_id, then creates new records.
```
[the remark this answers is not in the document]
pushyamig: Done as part of this https://github.com/tl-its-umich-edu/instructor-tools/pull/660/changes/fba49c053a091ca31d992b639e3f277c5a6c81f9

backend/canvas_app_explorer/alt_text_helper/alt_text_update.py
```diff
@@ -290,16 +291,16 @@ def delete_successfully_updated_items(self) -> None:
             logger.info(f"Deleted {deleted_count} successfully updated ImageItem records")
```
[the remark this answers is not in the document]
pushyamig: Done as part of this https://github.com/tl-its-umich-edu/instructor-tools/pull/660/changes/fba49c053a091ca31d992b639e3f277c5a6c81f9

backend/canvas_app_explorer/alt_text_helper/alt_text_update.py
```diff
@@ -290,16 +291,16 @@ def delete_successfully_updated_items(self) -> None:
...
                 # Check if this content still has any images
-                remaining_images = ImageItem.objects.filter(content_item__content_id=content_id).count()
+                remaining_images = ImageItem.objects.filter(content_item_id=content_item_id).count()
                 
                 if remaining_images == 0:
                     # No images left, safe to delete the ContentItem
-                    deleted_count, _ = ContentItem.objects.filter(content_id=content_id).delete()
-                    logger.info(f"Deleted orphaned ContentItem with content_id={content_id}")
+                    deleted_count, _ = ContentItem.objects.filter(id=content_item_id).delete()
+                    logger.info(f"Deleted orphaned ContentItem with id={content_item_id}")
```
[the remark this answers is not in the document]
pushyamig: Done as part of this https://github.com/tl-its-umich-edu/instructor-tools/pull/660/changes/fba49c053a091ca31d992b639e3f277c5a6c81f9

pushyamig:
I have to resolve conflicts from `main` due to some changes it brought in with `AltTextHome` and `AltTextReview`. I need to pass in the scan ID `AltTextReview` Figuring it out best way to do that.

pushyamig:
I am opening this for Review now. I have instruction on how to setup and test in the google sheet. Let me know if you want to talk via conversation or question, since this PR changes the existing design in some way

backend/canvas_app_explorer/alt_text_helper/views.py
```diff
@@ -207,6 +311,11 @@ def alt_text_update(self, request: Request) -> Response:
         try:
              # Extract unique content types from the payload
              content_types = list({item.get('content_type') for item in serializer.validated_data if item.get('content_type')})
+             
+             # Validate that all content and image IDs belong to the requesting course and content types
+             validation_error = self._validate_course_ownership(serializer.validated_data, course_id, content_types)
```
jaydonkrooss, line 317: I feel like this validation for an edge case is really late in the processing flow. We've never seen images outside of the course be requested, but if we did, shouldn't it be filtered out at the very beginning as a part of serializer validation?

Alternatively, the delete step could just filter down to content item id's that were actually successfully changed, since those will have been restricted to only content found in the course anyways. We'd likely have to create a new flag to `content_alt_text_update_report`  ( like `canvas_item_updated`?). Or we could coalesce it into the existing check for `is_alt_text_updated`, which only checks for errors right now.

backend/canvas_app_explorer/alt_text_helper/views.py
```diff
@@ -37,41 +38,58 @@ class AltTextScanViewSet(LoggingMixin, CourseIdRequiredMixin, viewsets.ViewSet):
...
     def start_scan(self, request: Request) -> Response:
         course_id = self._require_course_id(request)
+        obj = None
         
-        task_payload = {
-            'course_id': course_id,
-            'user_id': request.user.id,
-            'canvas_callback_url': request.build_absolute_uri(reverse('canvas-oauth-callback')),
-        }
         try:
```
jaydonkrooss, line 49: Nitpicky, but I feel like nearly all the comments in this try block are unnecessary and just repeat what code already conveys. If you rename `obj` to `newCourseScan`, I think you can just drop all the comments here. Maybe add a docstring to the method for a brief summary.

backend/canvas_app_explorer/alt_text_helper/views.py
```diff
@@ -37,41 +38,58 @@ class AltTextScanViewSet(LoggingMixin, CourseIdRequiredMixin, viewsets.ViewSet):
 
     def start_scan(self, request: Request) -> Response:
         course_id = self._require_course_id(request)
+        obj = None
```
jaydonkrooss: I know the variable `obj` had existed before your PR, but at this point with the model change it probably should have a more clear name like `newCourseScan`

backend/canvas_app_explorer/alt_text_helper/views.py
```diff
@@ -113,30 +131,111 @@ def __get_scan_course_content(self, course_id: int) -> object:
...
-        params = getattr(request, 'query_params', request.GET)
+        raw_params = getattr(request, 'query_params', request.GET)
+        params = raw_params.copy()
         serializer = ContentQuerySerializer(data=params)
         if not serializer.is_valid():
             logger.error("Invalid query parameters for get_content_images: %s", serializer.errors)
             return Response(status=HTTPStatus.BAD_REQUEST, data={"status_code": HTTPStatus.BAD_REQUEST, "message": serializer.errors})
 
         content_type = serializer.validated_data['content_type']
+        course_scan_id = serializer.validated_data.get('course_scan_id')
```
jaydonkrooss, line 212: I found a bug that if the user types a previous `course_scan_id` into the URL and submits, this code will unintentionally serve any existing unreviewed image items remaining from that older scan. 

My thought is that either we remove the scan id from the request + instead serve the "most recent" scan per course Id, or we make some kind of validation that the scan is the most recent. Seems like either way there will have to be a DB operation.

jaydonkrooss (changes requested):
I have a couple changes and a bug that would be good to fix for this PR. Other than that the model change isn't causing any glaring problems.

As I mentioned in slack, I am a little concerned about the state of the ContentItem/Image table if this brought into production and we don't "close the loop" on the review submission side. It might not be a major problem, but I'll create an issue describing my thoughts.

backend/canvas_app_explorer/alt_text_helper/views.py
pushyamig: So we can always fetch the latest scan and don't send the course_scan_id. 

But my concern was that if user had 2 scans `scan 1: Course 1` from one Tab, then possible in between scans they changed the course content images and did some `scan 2: Course 1` from different Tab,  The element list they show between will be different. 

If Scan 1 has 2 assignments ( Tab1) listed and Scan 2 has only 1 assignment(Tab2), then when we always fetch latest image content then if user operated from Scan 1 tab they will see only 1 assignment

Currently in the BE I am validating if the course scan id is associated with course id stored in the session, but there is no way to validate the user request on course scan id

I think we need come to a agree if we want to always fetch latest content based on latest scan after the showing scan results!

pushyamig: One other alternative we need to understand is currently the UI is handling Review Content page a Navigation, before it was loaded from `AltTextHome.tsx` before. So scan id was accessible AltTextReview.tsx as Props sent from `AltTextHome.tsx`. SO the bug now reported is not an issue

And now it is passing the `content_type` page or assignment as URL param and the code is reading this from URL `content_type`. So I stick this course_scan_id these. 
Either I need to figure out if I can pass the course_scan_id  to the AltTextReview.tsx as props. 

I will if that is possible otherwise we always need to fetch latest scan content for a course irrespective of what content summary listed on the screen for the user

backend/canvas_app_explorer/alt_text_helper/views.py
pushyamig: Done

backend/canvas_app_explorer/alt_text_helper/views.py
pushyamig: Done

backend/canvas_app_explorer/alt_text_helper/views.py
pushyamig: @jaydonkrooss , I moved away from sending the scan id from URL to React Router state, Let me know if this resolves your concern. I pushed the change now.

backend/canvas_app_explorer/alt_text_helper/views.py
pushyamig: >I feel like this validation for an edge case is really late in the processing flow. We've never seen images outside of the course be requested, but if we did, shouldn't it be filtered out at the very beginning as a part of serializer validation?

 Serializer verification can just validate if the Request payload has needed params like content/image ID. more or less structural of payload. It can't validate if data that is send is valid or not. 
 
 This check trying to filter if the possible ID are part of course scan content items or not. True FE won't send anything beyond what it know. This is extra layer of security. This is something Copilot suggested and I added upfront before we start process the Alt text label update. 
 
 So these are step of validation
 1. Serializer for Request Payload validation so we know exact item we need
 2. We want to make sure Content/Image Items sent was actually part the course

pushyamig: >Alternatively, the delete step could just filter down to content item id's that were actually successfully changed, since those will have been restricted to only content found in the course anyways. We'd likely have to create a new flag to content_alt_text_update_report ( like canvas_item_updated?). Or we could coalesce it into the existing check for is_alt_text_updated, which only checks for errors right now.

I know you raised concerns about this before. But the Simple Idea is every course scan table will have a column that say how many image items extracted with each scan like Course Scan 1: 26 image pulled. Once a request to update the alt text come we can delete those artifacts from respective tables, and If we want report on this in future we can definitely can get Based on the count on course scan and amount of images remained in the DB we can get that stats. 

You have asked this question before, should we keep all the content and don't delete upon alt text and you suggestion with  `new flag to content_alt_text_update_report ( like canvas_item_updated?)` is preparing to that. I like this insight on this we can discuss how this need to be done. But my challenge is to keep the scope to break the scan separate. 

I am still doing my high level analysis on reporting error, so we can talk if this want to understand this.

backend/canvas_app_explorer/alt_text_helper/views.py
```diff
@@ -36,42 +37,60 @@ class AltTextScanViewSet(LoggingMixin, CourseIdRequiredMixin, viewsets.ViewSet):
...
         except (DatabaseError, Exception) as e:
             message = f"Failed to initiate course scan due to {e}"
             logger.error(message)
+
+            if new_course_scan is not None:
+                new_course_scan.status = CourseScanStatus.FAILED.value
+                new_course_scan.save()
+                logger.error(f"Marked CourseScan {new_course_scan.id} as FAILED due to exception: {e}")
+
             return Response(status=HTTPStatus.INTERNAL_SERVER_ERROR, data={"status_code": HTTPStatus.INTERNAL_SERVER_ERROR, "message": message})
```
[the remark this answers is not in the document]
pushyamig: Done

backend/canvas_app_explorer/alt_text_helper/process_content_images.py
```diff
@@ -136,7 +140,7 @@ async def get_image_content_async(self, img_url):
             logger.error(f"HTTP error fetching image {img_url}: {http_err}")
             return http_err
         except Exception as req_err:
-            logger.error(f"Error fetching image content for image_id {img_url}: {req_err}")
+            logger.error(f"Error fetching image content for image_id {img_url}, course_scan_id {self.course_scan_id}, course_id: {self.course_id}: {req_err}")
```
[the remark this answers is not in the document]
pushyamig: Done

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -39,11 +39,12 @@
...
 def fetch_and_scan_course(task: Dict[str, Any]):
     logger.info(f"Starting fetch_and_scan_course for course_id: {task.get('course_id')}")
-      # mark CourseScan as running (create if missing)
+    # mark CourseScan as running
+    course_scan_id = int(task.get('course_scan_id'))
     course_id = int(task.get('course_id'))
 
     # adding a status before start of the scan, if this DB action failed no need to stop next steps of fetching content images
-    update_course_scan(course_id, CourseScanStatus.RUNNING)
+    update_course_scan(course_scan_id, CourseScanStatus.RUNNING, course_id=course_id)
```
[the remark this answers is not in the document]
pushyamig: Done

backend/canvas_app_explorer/alt_text_helper/views.py
jaydonkrooss: I see what you mean that we're currently using serialization to validate structure only. From my brief Copilot research, it seems to me technically possible for the serializer to do data validation as well, though it'd involve some different methodology we don't typically do: 

> PrimaryKeyRelatedField with a filtered queryset — You can scope an ID field to only accept valid DB records:
> ```
>content_item_id = serializers.PrimaryKeyRelatedField(
>    queryset=ContentItem.objects.all()  # or filtered per-course
>```

>Custom validate_<field>() or validate() methods — These run as part of .is_valid() and can query the DB to check ownership:
> ```
> def validate(self, data):
>     course_id = self.context['request'].course_id  # from session
>     if not ContentItem.objects.filter(
>         id=data['content_item_id'],
>         course_scan__course_id=course_id
>     ).exists():
> ```
> Context passing — The serializer can receive the course (from the session/view) via `context={'request': request}`, giving it everything needed to validate ownership.
> 

As to your second comment, I think you might have missed my suggestion was to do a filter so we wouldn't need validation at all. Successful content ids could be added to `alt_text_content_report`, so [this code](https://github.com/tl-its-umich-edu/instructor-tools/blob/a06907d8757a6b02e81d3de20942c834a8682142/backend/canvas_app_explorer/alt_text_helper/alt_text_update.py#L272-L279) would just automatically filters out any ids that didn't successfully submit to canvas. 

All this being said, I acknowledge this isn't blocking functionality, so I'm fine with leaving it. I also created #669 summarizes my thoughts on not deleting altogether, which is a separate concern that would probably replace much of this code anyways.

backend/canvas_app_explorer/alt_text_helper/views.py
jaydonkrooss: As I mentioned in our meeting, I'm happy with using react-router to hide the course_scan_id from URL. It is client-side storage so not completely tamper-proof; Ideally backend validation plus websockets to help with the stale data scenario you mentioned - I opened issue #670 so we can reference that idea later.

That being said, it works well with our current setup so I think this is resolved.

frontend/app/components/AltTextReview.tsx
```diff
@@ -75,26 +75,30 @@ export default function AltTextReview() {
...
+    };
+  }, [location.state]);
 
   useEffect(() => {
-    if (!categoryForReview) {
+    if (!categoryForReview || !scanIdFromUrl || scanIdFromUrl <= 0) {
       navigate('/alt-text-helper');
     }
-  }, [categoryForReview, navigate]);
+  }, [categoryForReview, scanIdFromUrl, navigate]);
```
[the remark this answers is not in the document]
jaydonkrooss: I think this is fine for now, especially because we've added the [exit guard](https://github.com/tl-its-umich-edu/instructor-tools/issues/567) if the user had changes before refresh. Perhaps we could change how this works in the future, but we'll need [websockets](https://github.com/tl-its-umich-edu/instructor-tools/issues/670) to help mitigate stale data.

jaydonkrooss (approved):
My comments are not blocking and would involve other issues - no need to address them here. Your changes are a good start to chunking out this much-needed work, thanks for taking it on!

The PR is ready to merge for me.

backend/canvas_app_explorer/alt_text_helper/views.py
pushyamig: Ok! Thanks for opening the issue!

backend/canvas_app_explorer/alt_text_helper/views.py
pushyamig: Oh... I see. i did not think this way. I will leave it for now. when I get a change next time I will change it. Thanks!

PR #664, opened 2026-04-14
Update copilot instructions for improvements around comment quality

jonespm:
No specific issue for this, was just noticing that sometimes when using AI coding assistants they didn't leave comments for future humans or AI for why things were done. I'd prompted for this but having this in the instructions may make it work automatically.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -29,10 +29,13 @@ A Django + React application for managing Canvas LTI tools and generating alt te
 ## Key Patterns & Conventions
 
 ### Views & Serializers
-- Use ViewSets with DRF's `LoggingMixin` to auto-log API requests with course context
-- Extract session data (`request.session['course_id']`, etc.) early; return 400 if missing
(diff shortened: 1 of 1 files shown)


jaydonkrooss (approved):
Interesting, I always found the comments from AI to be a little superfluous/wordy. I guess "non-obvious" is a good key-word to have. Perhaps this is more of a frontend thing, but often I'd just request it keep comments minimal, only adding when introducing new functions or entirely novel blocks of logic.

I'll try this method out more instead. Hopefully this makes it clear that comments should have an true goal beyond summarizing what can be easily gleaned.

jonespm:
I think without this I'd noticed it did a semi-confusing (to me) refactor of some code. I asked it to simply and make it clearer and also comment what was going on. I forget what specific issue this was on.

Since many of our projects we tend to not look at for 6 months+, coming back to them without having to figure out why something was done seemed like it would be quite useful. I guess we'll see what happens.

PR #683, opened 2026-05-20
issue_619 upgrade django-tinymce; default_link_target → link_default_target

zqian:
- upgrade django-tinymce to latest release tag
- updated default_link_target → link_default_target
- verified all setting options inside TINYMCE_DEFAULT_CONFIG


backend/settings.py
```diff
@@ -293,7 +293,6 @@
             ]
         }
     ],
-    "block_formats": "Paragraph=p",
```
jonespm, line 314: I'm not sure what this option was doing or why it was originally added, was this deprecated or just not needed?

jonespm (approved):
Seems fine for updating the link target. I'm not sure what [block_formats](https://www.tiny.cloud/docs/tinymce/latest/user-formatting-options/#block_formats) was doing or if that was causing issues too?

backend/settings.py
```diff
@@ -293,7 +293,6 @@
...
     "removed_menuitems": "fontformats fontsizes lineheight forecolor backcolor",
 }
```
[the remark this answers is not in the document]
zqian: added back the line `"block_formats": "Paragraph=p",`

backend/settings.py
zqian: I've added back the deleted line `"block_formats": "Paragraph=p",`

Issue #670, opened 2026-04-22
Consider websockets for Alt Text Helper


jaydonkrooss:
Alt Text Helper currently uses Tanstack Query and basic DRF endpoints to initiate course scans, poll for scan data, and submit changes. With more adoption and feature additions, we may see increased UI bugs or performance issues with the Rescan + review submission flow. Webhooks could be a solution to this similar to how they operate in ROHQ, though it may require some investigation.

This needs research: 1. determine performance limitations present in Alt Text Helper. 2. document the  endpoints + state relevant to rewrite scan initiate/ rescan, scan polling, and review submission. and 3.  determine if there are strengths/weaknesses to the library we bring in for this purpose

Issue #669, opened 2026-04-22
Review submission rework - keep content & images + error-flag


jaydonkrooss:
After deciding to keep records of past CourseScans #659 to better enable error-handling, we should additionally consider keeping Content + Image records upon submission. Needs research. Possible benefits: 
1. Error observability, we could populate the image/content item rows with failure reason and look back quickly
2. History feature to view past scans in the UI, so that there's a true record 
3. Re-using previously generated labels so that we don't have scaled AI costs with multiple scans.

For enabling error observability, the image or content record could be flagged during **either the scan or review** process and stored for future reference, either as a new item in an error table or as an attribute. 

Important considerations:
- there will have to be a migration effort, ideally to be done in conjunction with all error observability work 
- the "Skip" review action will have to be reworked or replaced with a different feature for users. Right now it removes items but keeping them makes it less permanent, so maybe we just flag it and intentionally copy the old record with a new scanId?  
- we may need to add an archiving process if the DB gets too bloated with duplicate records (depends on adoption and performance)

PR #687, opened 2026-05-26
I622 642 report errors

pushyamig:
#622 #659 

1. Partial Failure in content fetch should not stop from showcasing the image that has alt text 
2. I added a Debugpy workflow for Qworker, it works but you can't put on debugger on and edit the code. since that results in reload and won't work well need to restart the server. - I will this from this PR as I kept it will help review this PR
2. Tried to capture all error during the background scan and storing the DB
3. If Alt text generation fail, i mark it as failure but user can still go edit the alt text ( which would be empty text box)
4. If Image download fails, I don't present in the content  Preview list since user cannot see any image and alt text is empty
5. Model changes are there.  Should provide backward compatibility

**On canvas Dev look for Course 393, 394, 395,396, 397, 19 for some setup of failure cases**

**Suggest Docker build and starting of the server, No env change is required as defaults should work.** 

**Make sure Q worker should be 1 if you are configuring it via `.env` since Debugging the scan, debugger won't attach etc**

There is some debug code of failing content call, use it if you want it. Otherwise you can comeup with your own use cases for breaking this. I will delete all debug code.


frontend/app/components/AltTextHome.tsx
```diff
@@ -121,16 +125,64 @@ function AltTextHome(props: AltTextHomeProps) {
             </Box>
           ) : (
             <>
+              {scanErrors.length > 0 && (() => {
```
jaydonkrooss: This accordion should either be brought into LastScanInfo component or into it's own separate component, as a child to LastScanInfo. Either way we should also rename that file to `LastScanInfo.tsx` instead of `CourseScanComponent.tsx` to match its exported component name.

jaydonkrooss: If you do this I think it would be a better fit _beneath_ the Stack with the`Rescan Course` button , but we can also consult Janel on that decision

frontend/app/components/AltTextHome.tsx
```diff
@@ -121,16 +125,64 @@ function AltTextHome(props: AltTextHomeProps) {
...
+                return (
+                  <Accordion
+                    expanded={errorsExpanded}
+                    onChange={(_, expanded) => setErrorsExpanded(expanded)}
+                    sx={{ marginBottom: 2, border: '1px solid', borderColor }}
+                  >
+                    <AccordionSummary expandIcon={<ExpandMoreIcon />}>
+                      <Alert severity={severity} sx={{ width: '100%', padding: 0, background: 'transparent' }} icon={false}>
+                        <Typography variant='body2'>
+                          <strong>{summaryText}</strong> &mdash; {actionText}
```
jaydonkrooss: My nitpick is I don't like the HTML character entity (mdash, left/right quote on line 165) - I feel it's not as readable and unnecessary if we can write the characters `-` and `"` directly in JSX. 

Not a blocking problem though, if you feel otherwise

frontend/app/components/AltTextHome.tsx
```diff
@@ -121,16 +125,64 @@ function AltTextHome(props: AltTextHomeProps) {
...
+                      <Stack spacing={1}>
+                        {scanErrors.map((error, index) => (
+                          <Box key={error.id} sx={{ paddingLeft: 2 }}>
+                            <Typography variant='body2'>
+                              <strong>{index + 1}. {error.error_type}</strong>
+                              {error.error_title && <> ({error.error_title})</>}
+                              : {error.error_message}
+                            </Typography>
+                            {error.canvas_url && (
+                              <Typography variant='caption'>
```
jaydonkrooss: This `caption` variant is quite a small font, at least `body2` would be better. We could ask Janel as well, but I personally think the URL is probably more important to the user than other error details, maybe even `body1` would work here.

.vscode/launch.json
```diff
@@ -17,6 +17,22 @@
...
+                            }
+                    ],
+                    "connect": {
+                        "port": 5021,
+                        "host": "localhost"
+                    },
+                    "justMyCode": false
       }
   ]
 }
```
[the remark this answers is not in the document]
pushyamig: Done, I edited the file locally

.vscode/launch.json
```diff
@@ -17,6 +17,22 @@
             "host": "localhost",
         },
           "justMyCode": false
+            },
```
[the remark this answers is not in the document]
pushyamig: Done, edited the file locally instead  committing here

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -260,105 +444,198 @@ def get_assignments(course: Course):
...
+        return Quiz(source_quiz._requester, {
+            'id': random.randint(100000000, 999999999),
+            'title': random.choice(['Ghost Quiz', 'Phantom Exam', 'Broken Test', 'Mystery Quiz', 'Fake Assessment']),
+            'course_id': getattr(source_quiz, 'course_id', None),
+            'description': getattr(source_quiz, 'description', ''),
+        })
+
+    if quiz.id == 1277:  # specific quiz ID to trigger error for testing
+        quiz = _build_debug_quiz(quiz)
+
```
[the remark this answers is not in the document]
pushyamig: this is ok, I have let the reviewer know that there is some debug code so that it will help then in the review, this will be deleted before merging.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -20,10 +20,15 @@
 
 from backend import settings
 from backend.canvas_app_explorer.canvas_lti_manager.django_factory import DjangoCourseLtiManagerFactory
-from backend.canvas_app_explorer.canvas_lti_manager.exception import ImageContentExtractionException
-from backend.canvas_app_explorer.models import CourseScan, ContentItem, ImageItem, CourseScanStatus
+from backend.canvas_app_explorer.models import CourseScan, ContentItem, ImageItem, CourseScanStatus, CourseScanErrorLog
```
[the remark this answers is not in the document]
pushyamig: Done

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -56,39 +63,76 @@ def fetch_and_scan_course(task: Dict[str, Any]):
...
-            return
-
-        # Update that the course scan is completed
-        update_course_scan(course_scan_id, CourseScanStatus.COMPLETED, course_id=course_id)
+        content_fetch_result: Union[bool, List[CourseScanError]] = unpack_and_store_content_images(results, course, course_scan_id, course_id)
+
+        image_process_result: Union[bool, List[CourseScanError]] = retrieve_and_store_alt_text(course_scan_id, course_id, bearer_token=bearer_token)
+
+        # Determine final status: FAILED if both stages had errors, else COMPLETED
+        if content_fetch_result is not True or image_process_result is not True:
```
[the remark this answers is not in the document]
pushyamig: Done, edited this locally and pushed

docker-compose.yml
```diff
@@ -30,13 +30,17 @@ services:
...
     container_name: instructor_tools
     env_file:
       - .env
     environment:
       - DEBUG=True
       - RUN_FRONTEND=true
       - RUN_QWORKER_DEV_MODE=true
+      - QWORKER_DEBUGPY_ENABLE=true
+      - QWORKER_DEBUGPY_PORT=5021
+      - QWORKER_DEBUGPY_WAIT_FOR_CLIENT=false
```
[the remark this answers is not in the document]
pushyamig: This is ok, docker compose is only used for local development and Non-prod/prod instances are not affected by this. This will be more convenient rather changing it on needed basis and doing `docker compose up` each time

backend/settings.py
```diff
@@ -384,7 +384,7 @@
 #Django Q2 using default Django ORM
 Q_CLUSTER = {
     'name': os.getenv('Q_CLUSTER_NAME', 'IT_Cluster'),
-    'workers': int(os.getenv('Q_CLUSTER_WORKERS', 4)),
+    'workers': int(os.getenv('Q_CLUSTER_WORKERS', 1)),
     'timeout': int(os.getenv('Q_CLUSTER_TIMEOUT', 1800)), # 30 minutes
```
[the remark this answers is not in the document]
pushyamig: For Prod and Non-prod this `Q_CLUSTER_WORKERS ` value is exclusively set, so defaulting to 1 is ok.

docker-compose.yml
```diff
@@ -30,13 +30,17 @@ services:
     ports:
       - "5000:5000"
       - "5020:5020"
+      - "5021:5021"
```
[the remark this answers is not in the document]
pushyamig: I feel the risk is low on this.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -401,7 +678,7 @@ def _is_image_from_current_course(img_src: str, current_course_id: int) -> bool:
         return True
     except (ValueError, IndexError) as e:
         logger.warning(f"Could not parse course_id from URL {img_src}: {e}")
-        return False  # ignore image if we can't parse, to avoid losing valid images
+        raise e
```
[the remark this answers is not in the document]
pushyamig: Done, I think I was able to send the image URL as well along with raising the exception

backend/canvas_app_explorer/alt_text_helper/views.py
```diff
@@ -121,20 +125,83 @@ def __get_scan_course_content(self, course_scan_id: int) -> object:
...
-                        'image_count': ImageItem.objects.filter(content_item=content_item).count()
-                    }
-                    for content_item in content_queryset
-                ]
+                content_type_list = []
+                for content_item in content_queryset:
+                    image_count = ImageItem.objects.filter(
+                        content_item=content_item,
+                        image_process_state=ImageItem.IMAGE_STATE_SUCCESS,
+                    ).count()
```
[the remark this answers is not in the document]
pushyamig: Done, I have annotated as suggested.

backend/canvas_app_explorer/migrations/0030_imageitem_process_status.py
```diff
@@ -0,0 +1,30 @@
+from django.db import migrations, models
+
+
+def backfill_image_process_state(apps, schema_editor):
+    ImageItem = apps.get_model('canvas_app_explorer', 'ImageItem')
+
+    for image in ImageItem.objects.all().only('id', 'image_alt_text', 'image_process_state').iterator():
+        alt_text = image.image_alt_text or ''
+        image.image_process_state = 'success' if alt_text.strip() else 'failed'
+        image.save(update_fields=['image_process_state'])
```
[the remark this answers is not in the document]
pushyamig: This is ok, either way if all success or failed state, for Previously failed course just return [] for course_scan_error_details, this way the user is presented with success alt text images. 

The scope of user going to previous scanned course from a different term and doing an scan of all images might is thin. so this way only show good images, with out error  (which is current behavior)

```
{
    "found": true,
    "scan_details": {
        "id": 2,
        "course_id": 393,
        "status": "failed",
        "created_at": "2026-06-03T19:16:23.590610Z",
        "updated_at": "2026-06-03T19:16:29.279346Z",
        "total_image_count": 0,
        "course_content": {
            "assignment_list": [
                {
                    "id": 1,
                    "canvas_id": 4850,
                    "canvas_name": "Assignment - multiple(public)",
                    "image_count": 1
                },
                {
                    "id": 2,
                    "canvas_id": 4851,
                    "canvas_name": "Assignment - Canvas Public URL",
                    "image_count": 1
                }
            ],
            "page_list": [],
            "quiz_list": [],
            "quiz_question_list": []
        }
    },
    "scan_error_details": []
}
```

backend/canvas_app_explorer/alt_text_helper/background_tasks/error_logging.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    Persist one or more CourseScanError entries to the database with retry logic.
+    
+    Attempts to persist errors to the database with exponential backoff retries.
+    If all retries fail, errors are logged to console as a fallback mechanism.
+    This ensures that errors are not lost even if the database is temporarily unavailable.
+    
+    Retry behavior is controlled by module-level constants:
+    - ERROR_LOG_MAX_RETRY_ATTEMPTS: Maximum number of retry attempts
+    - ERROR_LOG_RETRY_BACKOFF_BASE: Base for exponential backoff (backoff = base^attempt seconds)
```
jaydonkrooss, line 25: Remove the constants description - you've already specified this in comments around the constant variables.

backend/canvas_app_explorer/alt_text_helper/views.py
```diff
@@ -119,22 +124,89 @@ def get_last_scan(self, request: Request) -> Response:
...
+
+    def _get_remediation_message(self, error_type: str, error_title: str | None) -> str:
+        """
+        Return a short user-facing remediation hint based on error classification.
+
+        Item-level errors (specific assignment/page/quiz/image failed) →
+            suggest editing or removing the image in Canvas.
+        System-level errors (whole fetch failed or infrastructure error) →
+            suggest retrying or contacting support.
+        Token errors → suggest refreshing browser or contacting support.
```
jaydonkrooss: This part of the description is kind of unnecessary, especially token vs system difference, which the code treats as the same thing anyways. Worth simplifying for readability in my opinion.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -112,99 +156,215 @@ async def get_courses_images(course: Course):
...
     )
-    images_with_alt_text = process_content_images.retrieve_images_with_alt_text()
-    return images_with_alt_text
-
-def unpack_and_store_content_images(results, course: Course, course_scan_id: int, course_id: int) -> bool:
-     # unpack results (assignments, pages) and handle exceptions returned by gather. gather maintain call order
+    image_process_state = process_content_images.retrieve_images_with_alt_text()
+    return image_process_state
+
+def unpack_and_store_content_images(
```
jaydonkrooss: This function has a lot of responsibilities now, plus the lengthy description is brittle to future changes + distracts from readability of the code itself. Instead I would lean on smaller comments in closer vicinity to the actual logic - ideally 3 line maximum.

One idea I have to simplify this method would be to only do the "unpacking" of the get_courses_images results, and return the successful content Items. Then save_scan_results is called separately without nesting it in here.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -112,99 +156,215 @@ async def get_courses_images(course: Course):
...
-        log_context = f"course_scan_id={course_scan_id}, course_id={obj.course_id}"
+        log_context = f"course_scan_id={course_scan_id}, course_id={course_id}"
         if status == CourseScanStatus.FAILED:
-            logger.error(error_message or f"Scan marked as FAILED for {log_context}")
+            # Construct error message from errors list
+            if errors:
+                error_summary = "; ".join([f"{e.get('type')}: {str(e.get('error'))}" for e in errors])
+                logger.error(f"Scan marked as FAILED for {log_context}. Errors: {error_summary}")
+                # Centralized error logging: persist errors to database
+                log_course_scan_errors(course_scan_id, errors)
```
jaydonkrooss, line 274: Thinking out loud here: passing a list `errors` as an argument for the purpose of inserting into `CourseScanErrorLog` doesn't feel like it should be a part of `update_course_scan`... A lot of code in this PR is returning mixed CourseScanError types to be carried downstream to insert here, but couldn't this be avoided if `log_course_scan_errors` were called at individual steps earlier? eg.  after `get_courses_images`, inside a catch block for `save_scan_results`, inside catch in `retrieve_and_store_alt_text`, etc. Then near the end of the background task, updating CourseScan to FAILED should be dependent on what's already in the ErrorLog, not the other way around.

I understand this would be a big overhaul of the working code and test cases, so I'm OK with keeping it as a separate issue, though I suspect could simplify a lot of the new code here.

backend/canvas_app_explorer/alt_text_helper/background_tasks/error_logging.py
```diff
@@ -0,0 +1,100 @@
...
+                for error in errors
+            ]
+            CourseScanErrorLog.objects.bulk_create(rows)
+            logger.debug(
+                "Logged %d CourseScanError entries to database: course_scan_id=%s",
+                len(rows),
+                course_scan_id,
+            )
+            return  # Success - exit early
+        except DatabaseError as db_error:
```
jaydonkrooss, line 53: I like the backoff functionality, eventually we should move it to a generic DB backoff decorator/custom function for other data operations throughout the app, with graceful error handling. 
Or we could bring in a package like [tenacity](https://github.com/jd/tenacity). Regardless, this should be created as a new issue.

backend/canvas_app_explorer/alt_text_helper/views.py
```diff
@@ -119,22 +124,89 @@ def get_last_scan(self, request: Request) -> Response:
...
+        """
+        try:
+            error_logs = CourseScanErrorLog.objects.filter(course_scan_id=course_scan_id).order_by('-created_at')
+            serializer = CourseScanErrorLogSerializer(error_logs, many=True)
+            return [
+                {
+                    'id': error.get('id'),
+                    'error_type': error.get('error_type'),
+                    'error_title': error.get('error_title'),
+                    'error_message': error.get('error_message'),
```
jaydonkrooss, line 191: I think it's been brought up that the full error message isn't necessarily useful to end users, and AI suggested to me that there could be accidental risk of exposing the verifier token for images & Canvas OAuth tokens for token failure. Probably unlikely/low risk, but to be safe I think its best to just remove it from the response and just give remediation message.

Otherwise we could try to sanitize, but it doesn't seem worth the additional effort.

backend/canvas_app_explorer/migrations/0030_imageitem_process_status.py
```diff
@@ -0,0 +1,30 @@
+from django.db import migrations, models
+
+
+def backfill_image_process_state(apps, schema_editor):
```
jaydonkrooss, line 4: Backfill is missing `total_image_count`, it might be important to add so not all previous failed scans don't accidentally flag "isCompleteError". Let me know what you think.

frontend/app/components/AltTextHome.tsx
```diff
@@ -121,16 +125,64 @@ function AltTextHome(props: AltTextHomeProps) {
...
+                const actionText = errorsExpanded
+                  ? 'collapse to hide details'
+                  : 'expand to view details';
+                return (
+                  <Accordion
+                    expanded={errorsExpanded}
+                    onChange={(_, expanded) => setErrorsExpanded(expanded)}
+                    sx={{ marginBottom: 2, border: '1px solid', borderColor }}
+                  >
+                    <AccordionSummary expandIcon={<ExpandMoreIcon />}>
```
jaydonkrooss: I got a couple of accessibility warnings from AI, but I feel these are not blocking this PR. Feel free to create a new issue to fix up these warnings; could also consult the a11y team if these are actually necessary.

- **`role="alert"` nested inside `role="button"`** — `<Alert>` inside `<AccordionSummary>`; screen readers double-announce/garble. Drop the `Alert`, style the summary directly, and move the live announcement out (below).
- **No live region** — when errors appear after polling, AT users aren't notified. Add an always-mounted `aria-live="polite"` status region with the summary text.
- **Numbered list is not semantic** — `Stack > Box > Typography` with `{index+1}.` text; use `<ol>/<li>` so AT can enumerate.
- **No `id`/`aria-controls`** linking summary↔details panel.
- **New-tab links** (`target="_blank"`) lack an "(opens in new tab)" cue (WCAG 3.2.5).
- **Warning border `#E2CF2A`** ≈ 1.57:1 on white, fails WCAG 1.4.11 (needs 3:1); it's the sole differentiator of partial vs complete severity → also a 1.4.1 use-of-color concern. Darken the token and/or add an explicit "Warning:"/"Error:" label.

frontend/app/components/AltTextHome.tsx
```diff
@@ -121,16 +125,64 @@ function AltTextHome(props: AltTextHomeProps) {
...
+                      <Alert severity={severity} sx={{ width: '100%', padding: 0, background: 'transparent' }} icon={false}>
+                        <Typography variant='body2'>
+                          <strong>{summaryText}</strong> &mdash; {actionText}
+                        </Typography>
+                      </Alert>
+                    </AccordionSummary>
+                    <AccordionDetails>
+                      <Stack spacing={1}>
+                        {scanErrors.map((error, index) => (
```
jaydonkrooss: Should we put a limit on this? In case of a lot of duplicate API errors for every alt text label, for example. If some get cut off, then perhaps a message of `# more problems, resolve the above and rescan` or some messaging like this would help. 

Not a blocking issue, but probably good for this release.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -56,39 +63,76 @@ def fetch_and_scan_course(task: Dict[str, Any]):
...
+                CourseScanStatus.FAILED,
+                course_id=course_id,
+                errors=merged_errors
+            )
+        else:
+            update_course_scan(course_scan_id, CourseScanStatus.COMPLETED, course_id=course_id)
+    except InvalidOAuthReturnError as e:
+        logger.error(f"OAuth token error in fetch_and_scan_course for course_id {course_id}: {e}")
+        oauth_error: CourseScanError = {
+            'type': 'Token Error',
```
jaydonkrooss: change to camelcase to keep consistent with other types
```suggestion
            'type': 'token_error',
```

backend/canvas_app_explorer/models.py
```diff
@@ -139,9 +150,60 @@ class ImageItem(models.Model):
...
+    
+    # FK to CourseScan primary key; when scan is deleted, errors are cascaded
+    course_scan = models.ForeignKey(
+        CourseScan,
+        on_delete=models.CASCADE,
+        db_column='course_scan_id',
+        related_name='scan_errors',
+    )
+    
+    # Type of error (e.g., 'content_database_save', 'image_process_error', 'alt_text_process_error', 'canvas_manager_setup_error')
```
jaydonkrooss: comments and tests mention this `canvas_manager_setup_error` type but it doesn't appear to be used. Either add the case to handle or remove the reference to this.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
```diff
@@ -260,105 +444,198 @@ def get_assignments(course: Course):
...
-        logger.error(f"Errors fetching Quizzes for course {course.id}: {e}")
-        raise e
 
-def process_quiz_with_questions(quiz: List[Dict[str, Any]], questions: List[Dict[str, Any]]):
+        mapped_quiz_question_results: List[List[Union[ContentItemWithImages, CourseScanError]]] = []
+        for quiz_obj, result in zip(quizzes, quiz_question_results):
+            if isinstance(result, Exception):
+                mapped_quiz_question_results.append([
+                    {
+                        'type': 'question',
```
jaydonkrooss: I'm having a hard time figuring out why `question` was added as a type different from `quiz_question`. This is referenced in the definition for CourseScanError and in the URL builder but I think it's not used for anything that `quiz_question` doesn't already do.

backend/canvas_app_explorer/alt_text_helper/background_tasks/types.py
```diff
@@ -0,0 +1,34 @@
+from typing import List, Optional, TypedDict, Union, NotRequired
+
+class CourseScanError(TypedDict):
+    """An image that could not be extracted due to an exception."""
+    type: str           # content type where the failure occurred: assignment | page | quiz | question | quiz_question
+    title: Optional[str]  # content title/context (e.g., assignment name, quiz name, or Course)
```
jaydonkrooss: The scope of this class seems to be more than just images. Edit the docstring and type examples since there are more than just content items that result in a CourseScanError.

Ideally a scan error's type should be a enum/constant set here and reused instead of using `frozenset` in `get_scan_course_content`. Or it could be baked into the model, similar to `ContentItem.CONTENT_TYPE_CHOICES`. That being said, we have a lot of scattered constants so I'm fine with leaving this as a separate issue.

jaydonkrooss (changes requested):
Nice work on handling a lot of areas of code. I've left a large batch of comments;  Some of them could become their own issue, or are non-blocking for the functionality that works. Essential changes I'd like to see are: 
- split up `unpack_and_store_content_images` responsibility
- move the accordion/banner code into its own React component
- remove the error_message from the UI, and 
- trim the large function docstrings & other quick format changes

In my local smoke testing, the app appeared to handle migration for old data fine and processed a handful of content failures well. I haven't yet gotten around to testing non-content failures, nor did I try debugging. I also haven't had time to look at the tests themselves, but they seem to pass fine.

frontend/app/components/AltTextHome.tsx
pushyamig: I am keeping it to `body2` since it doesn't look in larger font size then Error message itself. We can ask Janel if that needs to change

pushyamig: Done

backend/canvas_app_explorer/alt_text_helper/background_tasks/error_logging.py
pushyamig: Done

backend/canvas_app_explorer/alt_text_helper/views.py
pushyamig: ok, Done!

backend/canvas_app_explorer/alt_text_helper/background_tasks/error_logging.py
pushyamig: OK, I have created the issue. Don't take too much time to do this refactor. I just don't want to do too many big changes with this PR, if it is not blocking.

pushyamig: Here is the new issue: https://github.com/tl-its-umich-edu/instructor-tools/issues/693

backend/canvas_app_explorer/migrations/0030_imageitem_process_status.py
pushyamig: My Thinking to Backfilling on this....

I think backfilling `total_image_count` don't feel important, since this variable indicate how many image pulled in from a scan i.e content fetch and before image download and alt text generation happen. 

With past course scan there might most cases instructor done an action i.e approved or skipped a image alt text, that mean we would have deleted that image from our DB. even if we could count for backfilling either it count 0 or 2, whatever that number be but don't reflect the initial count what it was. so I felt this might be not unnecessary work for backfilling which might not be accurate Plus user might not go back to courses. a simple rescan would simply it. 

So my point at this time w.r.t to backfilling we don't need to create a workflow that might be **NOT** relevant for users, except for analytics ( which is something we want to do but don't have a concrete plan). Plus the error also not logged with this pilot release. so we have to make an exception if we do analytics

pushyamig: With that said, The UI aspect you pointed out for `total_image_count==0` for old scans the course scan error table doesn't exist so that would be empty, and status is displayed based on the status column in DB.

In the view we can give a `total_image_count` based on the number of image count from various content, but that still doesn't solve this problem showing any error since they don't exit. 

We can talk on this if you still have concerns.

frontend/app/components/AltTextHome.tsx
pushyamig: OK, what tools are you using for evaluating these warning or error for accessibility? this help me align with you and find solution

If it is not too much effort i will do it here.

frontend/app/components/AltTextHome.tsx
pushyamig: Done

frontend/app/components/AltTextHome.tsx
pushyamig: If you are saying
1. Error info to it's own component.... then I can do that
2, you are also stating that change the order of error accordion under the rescan button, when i show cased the work with error message to everybody, Janel did not see to give me the feedback on placement of error accordion.

pushyamig: OK, I renamed to LastScanInfo.tsx, now we have added a new file reference in this PR which we did not change.

frontend/app/components/AltTextHome.tsx
pushyamig: > could also consult the a11y team if these are actually necessary.

I don't know if we can consult them with a slack message, it might have to go through a review process.

frontend/app/components/AltTextHome.tsx
pushyamig: OK, I have ScanError.tsx component for all error details. It is now part of LastScanInfo.tsx

frontend/app/components/AltTextHome.tsx
pushyamig: Now we want to trying to interpret from the list of error coming our way. This should be handled in the backend and UI should just show the error not try to interpret, but we are trying to understand the type of error and it it is repetitive make it generic. I see this as Phase 2 based on the feedback received.   

Lots going on this PR and it might divert the scope of what this issue is trying to achieve.

frontend/app/components/AltTextHome.tsx
pushyamig: Done

frontend/app/components/AltTextHome.tsx
jaydonkrooss: > what tools are you using for evaluating these warning or error for accessibility?

I think it was Sonnet on Claude Code - not an actual accessibility tool specifically. I think the alert & aria-live region suggestion is not really necessary, but the other ones might be useful to add.

pushyamig: OK, I think I fixed the accessibility issue you mentioned is useful. I don't have Claude Code or Sonnet setup locally, So All i tested the functionality doesn't break due to the change. LMK if this is ok

pushyamig: Done

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
pushyamig: Done, I committed the change instead of accepting Apply suggestion here.

frontend/app/components/AltTextHome.tsx
jaydonkrooss: Sure, we can save the discussion for a future UI review issue. New component looks good.

frontend/app/components/AltTextHome.tsx
jaydonkrooss: Looks good, with local testing.

frontend/app/components/AltTextHome.tsx
jaydonkrooss: Sure. I'm fine with it being a separate issue.

backend/canvas_app_explorer/alt_text_helper/background_tasks/types.py
pushyamig: I agree, Content type was scattered and you could see the values are hard coded and not good. I created another issue to do this. https://github.com/tl-its-umich-edu/instructor-tools/issues/694

I also added more comment around CourseScanError type for clarity, LMK if that is reasonable!

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
pushyamig: Yes, I have created and issue to fix as a medium refactor https://github.com/tl-its-umich-edu/instructor-tools/issues/694.

For now that is supposed to be `quiz_question`, so i fixed it manually!

backend/canvas_app_explorer/models.py
pushyamig: Yes, I have corrected my comment for clarity now. Please see the commit: https://github.com/tl-its-umich-edu/instructor-tools/pull/687/changes/25da6685572faf2c55f498ab4b1fe5c6cd4bfc2e

frontend/app/components/AltTextHome.tsx
pushyamig: https://github.com/tl-its-umich-edu/instructor-tools/issues/694

backend/canvas_app_explorer/alt_text_helper/views.py
pushyamig: I feel this comment is counter intuitive to the PR I worked on, we wanted to show the reason for failure for user (grabbing error message from Exception), we really don't know why a failure occurs and don't want to understand and interpret what could be and rephrase something for privacy or security. Then i feel that workflow will be more work.  Also if when we state edit, fix, look the Course URL ( where the failure occurred) with out sharing a reason that will be as equal to not showing an error message.

Canvas token error doesn't send the token back in the error message it will state " Invalid Token" etc
Even with AI calls token is not sent back as error message, it sends the symptom only and we need to understand it.
I feel we are not showing instructor with verifier URL that is something he cannot see, since canvas is moving away from verifier token so we should not worry about this at this point. 

We can discuss this but I will defer not creating an issue as well.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
pushyamig: I had the course scan and error item call separate at first, But I have following could not resolve

1. We can only state a fetch failed/success after content fetch, image fetch finished. so that state if one step failed or otherwise that stated needs to be carried until the last. So I just held the error object instead of state of failure

2. If saving error items at each level say assignment, page image etc then I don't know if that call fails ( DB save) what to do. so what I did is i collected all the error list and made a single DB call, and added retry logic for extra safety. 

Before you see this design, I need to catch all the exception that were that was one deep. regardless if those error happens or not. Like I need to differentiate if image processing error as caused at image or alt text level so the action is different for both.  Also hook up a debugger so i have to solve 2 step before I come here.... 

we can always perfect this.... but not in this PR.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
pushyamig: I will look into this tomorrow... It might look lot of lines of code, all it is doing it unpack and save. 

I have to see how the error flow would effect if I break.

pushyamig: OK, I refactored as requested https://github.com/tl-its-umich-edu/instructor-tools/pull/687/changes/0cc6c25405a468585221955bd0401a82e4a43e86

backend/canvas_app_explorer/migrations/0030_imageitem_process_status.py
jaydonkrooss: Ah you're right, I forgot that the scan error table would be empty for old scans. Your other point make sense too, and we'd encourage users to rescan anyways.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
jaydonkrooss: I understand the challenges you mentioned, ultimately I'm fine with this being the design we go with for now. Here are my last thoughts to refer to next time:

Overall my instinct is to lean on the database as the source of truth for errors as much as possible, which _should_ address your first point: ie.`type` on CourseScanErrors should supply all the info we need to declare an entire scan failure or success, which is already done on this PR in the polling response. Deferring inserting all errors to the end of a scan shouldn't be strictly necessary for this.

I acknowledge your second point of database failures is tricky, but multiple log insert points wouldn't make it more unsafe in my opinion, just noisy on the console logs. Another solution in that case ought to be a websocket on the frontend, which notifies the user if an error-log insert fails. Touches #693 and #670

Some of this work could even be brought into #669, or any other API refactor issue we come up with in the future. It may take more time and planning than we have to spare right now.

pushyamig: IMHO, those decision we need to talked before me trying to put in all in one big PR. 

I tried to be as transparent as possible with my design and what I was thinking. So these are things we could have talked before asked right question and correct my thinking so that the PR is upto everybody's understanding. 

With that said, I really don't feel we discuss those thing in details from my past working experience. we mostly talk about requirements, what workflow solves the problem like using Django-q or DRF or D3 etc. I personally feel this is problem with small team working in multiple project and there is no said design principle we follow or improve upon design. Mostly we build something and work on 2 years on medium size feature and put the App in maintenance mode.

backend/canvas_app_explorer/alt_text_helper/views.py
jaydonkrooss: I'll resolve this for now, but my position is not against the purpose of error logging - you're right to say keeping `error_message` will be useful for understanding the error, especially for developers to diagnose new problems that arise. However, exposing it to UI isn't best practice because we can't control what might be included in it from external providers and tools.

Beyond the exposure risk (not a problem now, but possible), this also poses a UX problem of too many instructions: steering the user to OUR remediation should take priority over an Exception printout, which can be distracting if too long or even give bad information, eg. this real image problem includes two non-helpful URLs:

```
image_process_error (problem page): Client error '404 Not Found' for url 'https://www.oreilly.com/library/view/advanced-mysql-8/9781788834445/assets/test.png' 
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
View "problem page" in Canvas - Edit or delete the image in this content
```

Best case scenario the instructor just ignores the first two lines, but the worst case is they actually open the oreilly URL, or even the mozilla URL, instead of the true remediation. I will defer to the group on this, but I suspect that instructors will see it as noise more times than not.

backend/canvas_app_explorer/models.py
jaydonkrooss: I still see it leftover in the tests, once cleaned up I think this looks good.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
jaydonkrooss: I think it'd be good to rename this function to just unpack_content_images, and rephrase the description to mention that it's extracting errors out of the canvas response. It's no longer saving images

jaydonkrooss (approved):
Thanks for addressing it quickly, I've resolved most of the comments. Just one more quick cleanup suggestion on the `unpack_and_store_content_images` thread but other than that, this code looks ready to merge. Hopefully we'll have more time to look at the spinoff issues in the future, but not necessarily needed for this release.

backend/canvas_app_explorer/alt_text_helper/background_tasks/canvas_tools_alt_text_scan.py
pushyamig: Done, I fixed bunch of tests associated with this

PR #700, opened 2026-06-11
#691 Mysql support for SSL

pushyamig:
Fixes #691 

Local Optionally support 'DB_SSL_MODE' and DB_SSL_CA, Both values are optional in .env

To Test: If configuration is working
1.  In `Setting.py` Set `os.getenv('DB_SSL_MODE', 'REQUIRED') --->  os.getenv('DB_SSL_MODE', 'SOMETHING_RANDOME')` something random string and do a `docker compose up`, this will throw an error, strong indication that, wiring is working
2.  In `Setting.py` Set `os.getenv('DB_SSL_MODE', 'VERIFY_CA')`, but you did not provide any SSL CA file (.pem file) then it will throw an error, this is to check saying our connection is working as planned

Smoke Test: By default app launch files and DB inserts are happeing

### DIFF
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -117,9 +117,10 @@ def _get_db_options():
     """
     options = {'charset': 'utf8mb4'}
     options['ssl_mode'] = os.getenv('DB_SSL_MODE', 'REQUIRED')
-    
+
     ca_cert = os.getenv('DB_SSL_CA')
-    ca_cert and options.update({'ssl_ca': ca_cert})
+    if ca_cert:
+        options['ssl'] = {'ca': ca_cert}
     
     return options
 
--- a/backend/settings.py
+++ b/backend/settings.py
@@ -112,7 +112,7 @@ def _get_db_options():
     Build database OPTIONS dict with conditional SSL support.
     
     SSL Configuration:
-    - DB_SSL_MODE: Set to 'REQUIRED', 'VERIFY_IDENTITY', 'VERIFY_CA', or 'DISABLED' (default: 'REQUIRED')
+    - DB_SSL_MODE: Values include 'REQUIRED', 'VERIFY_IDENTITY', 'VERIFY_CA', 'DISABLED', 'PREFERRED'
     - DB_SSL_CA: Path to CA certificate (optional, used for verification)
     """
     options = {'charset': 'utf8mb4'}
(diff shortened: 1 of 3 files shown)


backend/settings.py
```diff
@@ -106,6 +106,23 @@
...
 # https://docs.djangoproject.com/en/3.2/ref/settings/#databases
 
+# Helper function to build database OPTIONS with optional SSL support
+def _get_db_options():
+    """
+    Build database OPTIONS dict with conditional SSL support.
+    
+    SSL Configuration:
+    - DB_SSL_MODE: Set to 'REQUIRED', 'VERIFY_IDENTITY', 'VERIFY_CA', or 'DISABLED' (default: 'REQUIRED')
+    - DB_SSL_CA: Path to CA certificate (optional, used for verification)
```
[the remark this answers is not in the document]
pushyamig: Done

backend/settings.py
```diff
@@ -106,6 +106,24 @@
...
+def _get_db_options():
+    """
+    Build database OPTIONS dict with conditional SSL support.
+    
+    SSL Configuration:
+    - DB_SSL_MODE: Values include 'REQUIRED', 'VERIFY_IDENTITY', 'VERIFY_CA', 'DISABLED', 'PREFERRED'
+    - DB_SSL_CA: Path to CA certificate (optional, used for verification)
+    """
+    options = {'charset': 'utf8mb4'}
+    options['ssl_mode'] = os.getenv('DB_SSL_MODE', 'REQUIRED')
```
jonespm: Does this still work with the local database? It looks like REQUIRED wouldn't fall back to no SSL? It looks like this new configuration sets ssl_mode even if there's no ssl cert provided? 

I think this should be refactored so that `ssl_mode` is only set if ca_cert is provided so it should be in the if statement below.

pushyamig: >Does this still work with the local database?
locally this seems to be fine

>It looks like REQUIRED wouldn't fall back to no SSL? 
Yes, U believe so but that is for Local, felt we don't have to worry about it.

Now I made the change as such if CA file is available then I am setting to VERITY_CA. Locally the .pem file provided won't work anyway since it throws as `TLS/SSL error: self-signed certificate in certificate chain` since AWS RDS self sign it local mysql setup doesn't find the Root associated. So `VERITY_CA` should be ok? since we will be providing this file to Non-prod/prod instances and not local

jonespm (approved):

PR #702, opened 2026-06-18
#691 ssl_mode required is default

pushyamig:
Fixes #691 

Make the REQUIRED as default since we only have CA cert and VERIFY_* mode won't work with it


zqian (approved):
Make sense!

